PR #61, opened 2021-04-09
[#60] Throwing problems with http headers

### LINKED ISSUE #60: Throwing problems with http headers
Zalando library does not provide an easy way to define custom headers when throwing `ThrowableProblem` - problem itself describes only response body. This becomes a challenge, when developer wants to have both - `application/problem+json` and custom headers (e.g for http 429 + `Retry After` header).

One possible walk around is to throw `WebApplicationException` with `Response` where we can set headers and provide `Problem` as entity/body:
```
throw new WebApplicationException(
    Response
        .status(Response.Status.BAD_REQUEST)
        .header("X-Special-Header", "ABC")
        .type("application/problem+json")
        .entity(
            Problem.builder()
                .withStatus(Status.BAD_REQUEST)
                .withTitle("Problem title")
                .with("custom_field", "CustomValue")
                .build()
            )
        .build()
);
```

This works pretty well, but such `WebApplicationException` will skip all ExceptionMappers and our PostProcessors, so it will not be enhanced (`instance` default value, mdc fields injection) nor logged, so does not provide consistent developer experience.

Provide new abstraction `HttpProblem` + nice builder + ExceptionMapper that will handle this type properly. Possible usage:
<pre>
throw HttpProblem.builder()
    .withStatus(Status.BAD_REQUEST)
    .withTitle("Problem title")
    <b>.withHeader("X-Special-Header", "ABC")</b>
    .with("custom_field", "CustomValue")
    .build();
</pre>

Zalando's `ProblemBuilder` class is final, so we have to create our own copy with headers functionality.

This should produce exactly the same response than WebApplicationException in the example above, but transformed by PostProcessors (mdc fields, logging etc):
```
HTTP/1.1 400 Bad Request
X-Special-Header: ABC
Content-Type: application/problem+json
        
{
  "title": "Problem title",
  "status": 400,
  "custom_field": "CustomValue",
  "mdc_field": "123"
}
```

This could also be a step towards getting rid of Zalando lib dependency.

lwitkowski:
See #60 for details and rationale. HttpProblem will probably slightly change in the next PRs (maybe we will decide to extract interface as Zalando did?).

Closes #60

### DIFF
--- a/runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
+++ b/runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
@@ -1,9 +1,12 @@
 package com.tietoevry.quarkus.resteasy.problem;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.net.URI;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.zalando.problem.Status;
 
 class HttpProblemTest {
@@ -39,6 +42,13 @@ class HttpProblemTest {
 
         assertThat(copy)
                 .isEqualToComparingFieldByField(original);
+    }
+
+    @ParameterizedTest()
+    @ValueSource(strings = { "type", "instance", "detail", "status" })
+    void builderShouldNotAllowAddingReservedProperties(String property) {
+        assertThatThrownBy(() -> HttpProblem.builder().with(property, "not relevant"))
+                .isInstanceOf(IllegalArgumentException.class);
     }
 
     private HttpProblem sampleHttpProblem() {
--- a/runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
+++ b/runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
@@ -1,0 +1,56 @@
+package com.tietoevry.quarkus.resteasy.problem;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.net.URI;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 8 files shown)


runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Test
+    void copyBuilderShouldCopyAllFields() {
+        HttpProblem original = sampleHttpProblem();
+
+        HttpProblem copy = HttpProblem.builder(original).build();
+
+        assertThat(copy)
+                .isEqualToComparingFieldByField(original);
```
pazkooda, line 41: `.isEqualToComparingFieldByField` is deprecated?

runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
```diff
@@ -0,0 +1,56 @@
+package com.tietoevry.quarkus.resteasy.problem;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.net.URI;
+import org.junit.jupiter.api.Test;
+import org.zalando.problem.Status;
+
+class HttpProblemTest {
```
pazkooda, line 9: Possibly add test on reserved properties throwing `IllegalArgExc`.

pazkooda (approved):
LGTM (with comments 🙂 )

runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
lwitkowski: Nope, it's not, at least not in assertj version we're using.

runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/HttpProblemTest.java
lwitkowski: Good point, adding now.

PR #62, opened 2021-04-09
Remove dependency on Zalando Problem library

lwitkowski:
Main changes: 
`Problem` interface from Zalando Lib is no longer a core class used everywhere in the extension - it is replaced by `HttpProblem`, which also carries http headers, and have nice 'clone builder' that helps building `HttpProblem`s based on other `HttpProblem`s.
`Status` from Zalando's lib got replaced with `Response.Status` from JaxRS everywhere as well.

Thanks to `HttpProblem` having place for http headers it was super easy to simplify `WebApplicationExceptionMapper` so that it extends `ExceptionMapperBase` again, but without extending api surface of this base class.

`ProblemUtils` class is gone.

Zalando's Problem is supported though (we provide a mapper if library is in the classpath).

This implementation is 4-5 times faster than the one based on Zalando's Problem :)


integration-test/src/test/java/com/tietoevry/quarkus/resteasy/problem/JaxRsMappersIT.java
```diff
@@ -50,23 +48,11 @@ void webApplicationExceptionShouldReturnHeaders() {
                 .log().all()
                 .statusCode(status)
                 .header(HttpHeaders.RETRY_AFTER, equalTo("120"))
-                .body("title", equalTo(Status.valueOf(status).getReasonPhrase()))
+                .body("title", equalTo(Response.Status.fromStatusCode(status).getReasonPhrase()))
                 .body("status", equalTo(status))
                 .body("stacktrace", nullValue());
     }
 
-    @Test
```
pazkooda, line 58: Why it is gone?

runtime/pom.xml
```diff
@@ -40,12 +46,7 @@
             <groupId>org.zalando</groupId>
             <artifactId>problem</artifactId>
             <version>${zalando-problem.version}</version>
-        </dependency>
-        <!-- fix missing javax.annotation.concurrent.Immutable -->
-        <dependency>
-            <groupId>com.google.code.findbugs</groupId>
-            <artifactId>jsr305</artifactId>
-            <version>3.0.2</version>
+            <optional>true</optional>
```
pazkooda, line 49: 👍

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapper.java
```diff
@@ -16,8 +21,36 @@
...
+                .build();
+    }
+
+    private Violation toViolation(ConstraintViolation<?> constraintViolation) {
+        return new Violation(
+                constraintViolation.getMessage(),
+                dropFirstTwoPathElements(constraintViolation.getPropertyPath()));
+    }
+
+    private String dropFirstTwoPathElements(Path propertyPath) {
```
pazkooda, line 43: Good time to rename it to purpose of it. My suggestion:

```java
dropMethodNameAndArgumentPostionFromPath(...)
```

pazkooda (approved):
LGTM, see details ;)

integration-test/src/test/java/com/tietoevry/quarkus/resteasy/problem/JaxRsMappersIT.java
lwitkowski: It was moved to separate ZalandoProblemMapperIT to make zalando stuff even more separated from everything.

PR #64, opened 2021-04-12
Opening HttpProblem for inheritance

lwitkowski:
I see a need to be able to create own problem classes (e.g in jHipster quarkus project), so that business code does not have to use builder() everywhere, and just throw nice `OutOfStock` problem instead.

This is a proposal that I personally find good/easy enough.

### DIFF
--- a/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
+++ b/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
@@ -2,17 +2,18 @@ package com.tietoevry.quarkus.resteasy.problem;
 
 import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
 
-import io.smallrye.common.constraint.Nullable;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -21,7 +22,7 @@ import javax.ws.rs.core.Response;
  * Representation of RFC7807 Problem schema.
  */
 @Immutable
-public final class HttpProblem extends RuntimeException {
+public class HttpProblem extends RuntimeException {
 
     public static final MediaType MEDIA_TYPE = new MediaType("application", "problem+json");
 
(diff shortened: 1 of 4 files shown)


runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
```diff
@@ -21,7 +22,7 @@
  * Representation of RFC7807 Problem schema.
  */
 @Immutable
-public final class HttpProblem extends RuntimeException {
+public class HttpProblem extends RuntimeException {
```
pazkooda, line 25: There are 2 things here that I'd like to express.

First of all why only `RuntimeException`?
Second: I'd prefer one of 2 options: 
* leave this class `final` or
*  *properly* extract interface and provide abstract class for convenience with some default implementation based on it - similar to Zalando lib.
Having this open (extendable) now will close possibility to avoid backward compatible changes in future - someone nearly for sure gonna extend this class it its code.

pazkooda (commented):
This is library. Such unsafe extension point is asking yourself for a trouble in future.
Aka _Effective Java Item 17_: 
> Design and document for inheritance or else prohibit it

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
lwitkowski: I kind of agree on principle, but I personally don't see any benefit of such abstract class: can you elaborate on this? Why having abstract class is supposed to be easier to maintain? Once we need to make breaking changes for our extension we still can switch to using internal model (create `InternalProblem` class) without changing HttpProblem class public interface. Lets analyse all pros and cons of complicating types hierarchy.

Rule no 1 for me: let's keep it simple. For our needs I personally think we can go with a bit simple (maybe even 1 class) setup as is presented now. Zalando classes all depend on each other - interface has factory methods creating specific implementation instance etc - I'm not sure if I'm in fond of it too much :)

lwitkowski: Regarding `RuntimeException`: why not? ;) why `WebApplicationException extends RuntimeException`?

lwitkowski:
> This is library. Such unsafe extension point is asking yourself for a trouble in future.
> Design and document for inheritance or else prohibit it

Design - checked (immutability, encapsulation). Documentation - checked. What makes you think it's unsafe?

pazkooda (approved):
As discussed in-person: since our `HttpProblem` is already more then _RFC7807 Problem_ than we can consider it a "base class for extension" at `Resource`s level.

Yet to make throwing Problems more versatile/convenient we may need to consider extract _pure_ `Problem` (most likely in form of interface) so - when applicable - this can also be thrown from business layer code.

To be worked on in next iterations :)

lwitkowski:
> To be worked on in next iterations :)

Amen, thanks for good discussion!

Issue #56, opened 2021-04-02
Enable Dependabot


lwitkowski:

lwitkowski:
Good stuff @Wertros, closing this one.

PR #71, opened 2021-04-15
Quarkus bumped to 1.13.2.Final

lwitkowski:


run-jvm-tests
```diff
@@ -4,7 +4,7 @@ if [[ "$java_ver" != "8" ]]; then
     exit;
 fi
 
-./mvnw clean install -DskipTests
+./mvnw clean install -Pquick
```
pazkooda, line 7: ?

pazkooda: Have you forgotten to push `quick` profile or this is magic I'm not aware of?

lwitkowski: `?` is very ambigous comment, let's try to avoid it, as it can mean anything (what this change is all about? why this change is this PR? does this `quick` profile exist already?) :)

`quick` profile (custom one) turns off the tests, formatting and forbiddenapi checks - introduced some time ago already.

I run those scripts to check bumping dependency versions locally, they are not used outside of this context.

lwitkowski: `quick` is used in CI builds for some time now, that's the conclusion :)

lwitkowski:
On the other hand - this could be nice test case for dependabot config #56

run-jvm-tests
Wertros: If `quick` is a profile then you could also write it in this way `./mvnw clean install -P quick` and it's much better to understand IMHO

lwitkowski: I agree, it's good candidate for separate chore PR (script files, GA workflows, documentation). For the sake of consistency I'd keep it as is for now.

Wertros (approved):
LGTM

run-jvm-tests
pazkooda: I was not aware that we already have such profile somewhere in `pom.xml`'s. Thus you got me by surprise. I like this one - glad to learn about it 🙂

pazkooda: BTW @Wertros, I've never ever used `-P` with space so it is way much more natural not to have space. 
Most likely it is from times when we used to use `-Dparam.here` in plain Java times.

lwitkowski: For the record: https://maven.apache.org/guides/introduction/introduction-to-profiles.html#details-on-profile-activation
Spoiler: official maven docs use spaces :) Quarkus does not use spaces in their CI. I personally don't mind, both versions are fine by me.

Wertros: @pazkooda from maven official documentation `mvn groupId:artifactId:goal -P profile-1,profile-2,?profile-3`
I am not saying your solution is wrong but it's just my opinion that is much more readable having space between.

PR #74, opened 2021-04-16
Build-time ProblemProcessor simplified /with automatic classpath detection/ so it will be easier to add more mappers

lwitkowski:
`+31 −113` - isn't it what we all like? ;)


deployment/src/main/java/com/tietoevry/quarkus/resteasy/problem/deployment/ProblemProcessor.java
```diff
@@ -25,62 +25,59 @@
...
+
+        mappers.put("com.fasterxml.jackson.core.JsonProcessingException", "jackson.JsonProcessingExceptionMapper");
+        mappers.put("javax.ws.rs.ProcessingException", "jsonb.RestEasyClassicJsonbExceptionMapper");
+        mappers.put("javax.json.bind.JsonbException", "jsonb.JsonbExceptionMapper");
+
+        return mappers.entrySet().stream()
+                .filter(mapper -> {
+                    String exceptionClass = mapper.getKey();
+                    return classAvailable(exceptionClass);
+                })
```
Wertros, line 52: Maybe it's better to extract whole logic of lambda and put it to one method called `isClassAvailable()` or `hasRegisteredClass()` or `isOnClasspath()`
For me it's always better to have a prefix for all boolean methods i.e `is` or `has` ...

deployment/src/main/java/com/tietoevry/quarkus/resteasy/problem/deployment/ProblemProcessor.java
```diff
@@ -25,62 +25,59 @@
     private static final String FEATURE_NAME = "resteasy-problem";
     private static final String PACKAGE = "com.tietoevry.quarkus.resteasy.problem.";
```
Wertros, line 26: Maybe `INTERNAL_PACKAGE`

Wertros (approved):
`+31 −113` - isn't it what we all like? - TRUE  👍

deployment/src/main/java/com/tietoevry/quarkus/resteasy/problem/deployment/ProblemProcessor.java
lwitkowski: Agree, `PACKAGE` is not very descriptive. 
Another option: `EXTENSION_MAIN_PACKAGE` ?

deployment/src/main/java/com/tietoevry/quarkus/resteasy/problem/deployment/ProblemProcessor.java
lwitkowski: I can add `is`, no problem. Remember though, that this method accepts either exception class name, or not-very-readable `Map.Entry`, as we're filtering map, not a list. 

Thus:`filter` accepts `Predicate<Map.Entry>`, not `Predicate<String>`. I find it more readable extracting entry.getKey into properly named variable first (`exceptionClass`), then passing it to final method.

PR #97, opened 2021-04-20
Quickstart workflow with jackson and jsonb (failing) runner

lwitkowski:


.github/workflows/quickstart-tests.yaml
```diff
@@ -27,15 +30,23 @@ jobs:
...
 
-          mvn io.quarkus:quarkus-maven-plugin:1.13.2.Final:create -Dextensions="resteasy,resteasy-jsonb" -DprojectGroupId=problem -DprojectArtifactId=quarkus-resteasy-problem-playground -DclassName="problem.HelloResource" -Dpath="/hello"
-          sed 's/<dependencies>/<dependencies><dependency><groupId>com.tietoevry.quarkus<\/groupId><artifactId>quarkus-resteasy-problem<\/artifactId><version>'"$EXTENSION_VERSION"'<\/version><\/dependency>/g' quarkus-resteasy-problem-playground/pom.xml > quarkus-resteasy-problem-playground/pom2.xml
+      - name: Add extension dependency
+        working-directory: ./quickstart/quarkus-resteasy-problem-playground
+        run: |
+          EXTENSION_VERSION=$(./mvnw -f ../../pom.xml -q -Dexec.executable=echo -Dexec.args='${project.version}' --non-recursive exec:exec)
+          sed 's/<dependencies>/<dependencies><dependency><groupId>com.tietoevry.quarkus<\/groupId><artifactId>quarkus-resteasy-problem<\/artifactId><version>'"$EXTENSION_VERSION"'<\/version><\/dependency>/g' pom.xml > pom_new.xml
+          mv pom_new.xml pom.xml
```
pazkooda, line 47: `set -i` ?

lwitkowski: You mean this would allow sed to overwrite file in place? mv is exactly because sed was not overwriting original file...

pazkooda: Wrote this from the top of my head by as far as I remember `sed -i` does inline replace.

lwitkowski: Works like a charm, checked that on still-opened PR, thanks!

PR #103, opened 2021-04-24
[#73] [#102] Improve validation problem responses

### LINKED ISSUE #73: Failed validation of query param returns empty `field` in `violations` list
**Describe the bug**
Failed validation of primitive input returns empty 'field' in `violations` list

**To Reproduce**
Endpoint:
```
@GET
@Path("/search")
public String search(@QueryParam("phrase") @Length(min = 10) String phrase) {
    return "not relevant";
}
```
Request:
```
curl http://localhost:8080/search?phrase=abc
```

Response:
```
{
    (...)
    "violations": [
        {
            "field": "",
            "message": "length must be between 10 and 2147483647"
        }
    ]
}
```

**Expected behavior**
`field` in violation object should be `phrase` instead of empty string
```
{
    (...)
    "violations": [
        {
            "field": "phrase",
            "message": "length must be between 10 and 2147483647"
        }
    ]
}
```

### LINKED ISSUE #102: Improve validation problem responses
We see a need to improve ConstraintViolationException responses, as `field` values could be less cryptic for REST api client than they are now.

JaxRS endpoint example: 
```java
@POST
@Path("/resource/{param}")
public void endpoint(
        @Valid @PathParam("param") @Length(min = 100, max = 101) String pathParam,
        @Valid @QueryParam("param") @Length(min = 200, max = 201) String queryParam,
        @Valid @HeaderParam("param") @Length(min = 300, max = 301) String headerParam,
        @Valid RequestBody body) {
}

static class RequestBody {
    @Length(min = 6, max = 10)
    String param = "";

    @Valid
    List<Item> list = Collections.singletonList(new Item());

    static class Item {
        @Length(min = 7, max = 10)
        String param = "";
    }

    @Valid
    NestedObject nested = new NestedObject();

    static class NestedObject {
        @Length(min = 8, max = 10)
        String param = "";
    }
}
```

Current output that needs improvements (empty `field` for path, query and header param violations, lack of index for body `list.item`)
```json
{
    "status": 400,
    "title": "Bad Request",
    "instance": "/resource/1",
    "violations": [
        {
            "field": "",
            "message": "length must be between 100 and 101"
        },
        {
            "field": "",
            "message": "length must be between 200 and 201"
        },
        {
            "field": "",
            "message": "length must be between 300 and 301"
        },
        {
            "field": "param",
            "message": "length must be between 6 and 10"
        },
        {
            "field": "list.param",
            "message": "length must be between 7 and 10"
        },
        {
            "field": "nested.param",
            "message": "length must be between 8 and 10"
        }
    ]
}
```

Expected output (mind better `field` values and new property `in` in violation item):
```json
{
    "status": 400,
    "title": "Bad Request",
    "instance": "/resource/1",
    "violations": [
        {
            "field": "param",
            "in": "path",
            "message": "length must be between 100 and 101"
        },
        {
            "field": "param",
            "in": "query",
            "message": "length must be between 200 and 201"
        },
        {
            "field": "param",
            "in": "header",
            "message": "length must be between 300 and 301"
        },
        {
            "field": "param",
            "in": "body",
            "message": "length must be between 6 and 10"
        },
        {
            "field": "list[0].param",
            "in": "body",
            "message": "length must be between 7 and 10"
        },
        {
            "field": "nested.param",
            "in": "body",
            "message": "length must be between 8 and 10"
        }
    ]
}
```

To keep this description short @FormParam was not included, as it is mutually exclusive with json body payload, but it should be properly handled as well.

`field` values should also work properly if custom `propertyNodeNameProvider` is configured for Hibernate Validator.

Wertros:
- Fixed bug with empty field in violation (Closes #73 )
- Added new field into violations 'in' for clear understanding (Closes #102 )

### DIFF
--- a/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapper.java
+++ b/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapper.java
@@ -2,16 +2,12 @@ package com.tietoevry.quarkus.resteasy.problem.javax;
 
 import com.tietoevry.quarkus.resteasy.problem.ExceptionMapperBase;
 import com.tietoevry.quarkus.resteasy.problem.HttpProblem;
-import java.lang.reflect.Method;
-import java.lang.reflect.Parameter;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.stream.Collectors;
+
 import javax.annotation.Priority;
 import javax.validation.ConstraintViolation;
 import javax.validation.ConstraintViolationException;
 import javax.validation.Path;
+import javax.ws.rs.FormParam;
 import javax.ws.rs.HeaderParam;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Priorities;
@@ -19,6 +15,14 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.container.ResourceInfo;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Exception Mapper for ConstraintViolationException from Bean Validation API. Hibernate Validator, among others throw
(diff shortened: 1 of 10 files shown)


lwitkowski:
@Wertros Please have a look at #107 with some suggestions and additional test cases.

My suggestion for title of this PR (and final merge commit message): `[#73] [#102] Improve validation problem responses`

It's also good to add this:
```
Fixes #73
Fixes #102
```
as PR description, which will close related issues automatically after merging.

lwitkowski (changes requested):
@Wertros Please have a look at #107 with some suggestions and additional test cases.

My suggestion for title of this PR (and final merge commit message): [#73] [#102] Improve validation problem responses

It's also good to add this:

Fixes #73
Fixes #102
as PR description, which will close related issues automatically after merging.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapper.java
```diff
@@ -35,9 +45,41 @@ protected HttpProblem toProblem(ConstraintViolationException exception) {
...
-        return new Violation(
-                constraintViolation.getMessage(),
-                dropMethodNameAndArgumentPositionFromPath(constraintViolation.getPropertyPath()));
+        String invalidArg = extractArgValue(constraintViolation.getPropertyPath());
+        Parameter invalidParam = getInvalidParameter(invalidArg);
+        return createViolation(constraintViolation, invalidParam);
+    }
+
+    private Parameter getInvalidParameter(String invalidArg) {
+        int paramPosition = Integer.parseInt(invalidArg.substring(3));
```
pazkooda, line 54: Magic number `3`.

pazkooda (commented):
I like most of changes proposed by @lwitkowski in next PullReq, consider reviewing it and possibly merging.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapper.java
Wertros: You are right @pazkooda this can be named properly

lwitkowski: I did not comment on this topic, as it is not feasible solution in regards to custom naming provider - you cannot expect arg name to have digit/number at the end in this case. One possible approach is to look for parameters  by second segment of the path from ConstraintViolation - this will cover both cases I belive.

Wertros: That's also valid, I realized that after merging your PR into my branch and checked tests that are not passing.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/javax/Violation.java
```diff
@@ -1,29 +1,32 @@
...
+        query,
+        path,
+        header,
+        form,
+        body,
+        unknown;
 
-        public Violation violation(String message, String field) {
-            return new Violation(message, field, this);
+        public MessageSupplier violation(String field) {
```
pazkooda, line 13: Maybe I miss something but this is a bit over-engineered for me.

Wertros: Taking into consideration fact that this constructor takes arguments `String`, `String` it is very easy to provide wrong arguments in the wrong place which could lead to incorrect Violation object.
It's another piece of complexity to achieve simplicity :)

pazkooda: Builder?

lwitkowski: Yeap, it's an enhanced builder, which does not allow you to build an object that is broken/not finished. It will give you constructed object only if you provide all mandatory fields, which I found a great value for the price of more complicated builder implementation :)

I improved method names a bit (violation -> field, message -> withMessage()

lwitkowski (approved):
I applied small changes as commit - sorry for that, it was supposed to be PR targeting this PR, but my git did not create new branch as instructed). Nevertheless - it looks very nice IMO overall.

PR #107, opened 2021-04-25
Added unit tests for @FormParam, collection items validation and custom Hibernate propertyNodeNameProvider

lwitkowski:
This PR is a bit bigger suggestion to original PR: simplified Violation building  + mentioned tests for special cases - Disabled for now.

### DIFF
--- a/runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapperTest.java
+++ b/runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapperTest.java
@@ -2,8 +2,15 @@ package com.tietoevry.quarkus.resteasy.problem.javax;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.BeanDescription;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.introspect.BeanPropertyDefinition;
 import com.tietoevry.quarkus.resteasy.problem.HttpProblem;
 import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import javax.validation.ConstraintViolation;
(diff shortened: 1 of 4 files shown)


runtime/pom.xml
```diff
@@ -73,6 +73,12 @@
             <artifactId>quarkus-hibernate-validator</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.thoughtworks.paranamer</groupId>
+            <artifactId>paranamer</artifactId>
```
pazkooda, line 78: Maybe I'm not very eagle-eyed but I don't see any usages of this library in code.

runtime/src/test/java/com/tietoevry/quarkus/resteasy/problem/javax/ConstraintViolationExceptionMapperTest.java
```diff
@@ -20,63 +28,169 @@
...
+            return new StubResourceInfo(Validation.byProvider(HibernateValidator.class)
+                    .configure()
+                    .buildValidatorFactory()
+                    .getValidator());
+        }
+
+        public static StubResourceInfo withJsonPropertyAwareValidator() {
+            return new StubResourceInfo(Validation.byProvider(HibernateValidator.class)
+                    .configure()
+                    .parameterNameProvider(new ParanamerParameterNameProvider())
```
lwitkowski, line 168: Here @pazkooda ;) Hibernate already provides part of Paranamer as transitive dependency, but without explicitly adding it to pom.pl I get runtime exceptions (class not found)

pazkooda: Well, with package name starting with `org.hibernate.validator` and `groupId` being `com.thoughtworks.paranamer` might be difficult to spot in PullReq view.
Also it a bit strange that they "borrow" Java `package` from other library - can cause theoretical collisions in future 😕 .

lwitkowski: Hibernate can use paranamer library via `ParanamerParameterNameProvider` - it's an adapter from paranamer world to the Hibernate one :)

Wertros (approved):
I like your changes especially when calling i.e `Violation.In.Body` my intention was the same but yours seems to be even better 👍

PR #110, opened 2021-04-27
Sed replaced with quarkus:add-extension in Quickstarts Test workflow

lwitkowski:
You'll like this one @pazkooda :)


pazkooda (approved):
You are so wrong @lwitkowski,... I *love* it.
😜

PR #215, opened 2022-07-12
Use RFC-compliant URIs with scheme in instance field

### LINKED ISSUE #214: instance field may contain data that is not a URI
**Describe the bug**

**To Reproduce**
- Construct a new Quarkus project with the steps indicated in the project README
- `mvn quarkus:dev`
- `curl http://localhost:8080/unknown`

**Expected behavior**

The contents of the returned `instance` field are a valid URI conforming to https://www.rfc-editor.org/rfc/rfc3986#section-3 This is what is referred to from the problem+json RFC https://datatracker.ietf.org/doc/html/rfc7807#section-3.1.

Right now, the following JSON is returned:

```json
{
  "status": 404,
  "title": "Not Found",
  "detail": "RESTEASY003210: Could not find resource for full path: http://localhost:8080/unknown",
  "instance": "/unknown"
}
```

RFC 3986 always requires a scheme:

> The scheme and path components are required, though the path may be empty.

Therefore, instances should be prefixed with `http...`.

languitar:
Construct a real URI including a schema for the instance field to ensure
that the value conforms to the problem+json RFC.

This would fix #214

### DIFF
--- a/integration-test/src/test/java/com/tietoevry/quarkus/resteasy/problem/GenericMappersIT.java
+++ b/integration-test/src/test/java/com/tietoevry/quarkus/resteasy/problem/GenericMappersIT.java
@@ -26,7 +26,7 @@ class GenericMappersIT {
                 .get("/throw/generic/runtime-exception")
                 .then()
                 .statusCode(INTERNAL_SERVER_ERROR.getStatusCode())
-                .body("instance", equalTo("/throw/generic/runtime-exception"))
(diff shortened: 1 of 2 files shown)


lwitkowski:
Hi @languitar, are you planning to move this PR forward, or we should rather close it?

languitar:
Sorry, I didn't find the time to adapt this and now I am not using Quarkus actively anymore.

lwitkowski:
Thanks for info, I'll close it for now.

PR #238, opened 2022-10-20
[#237] Fix for: HttpAuthenticator is ignored when authentication fails

### LINKED ISSUE #108: Missing `WWW-Authenticate` header when `UnauthorizedException` or `AuthenticationFailedException` is thrown
**Describe the bug**
From [HTTP RFC](https://tools.ietf.org/html/rfc7235#section-4.1):
```
 A server generating a 401 (Unauthorized) response MUST send a
 WWW-Authenticate header field containing at least one challenge.  A
 server MAY generate a WWW-Authenticate header field in other response
 messages to indicate that supplying credentials (or different
 credentials) might affect the response.
```
Default Quarkus response for these exceptions contains `WWW-Authenticate` header:
```
HTTP/1.1 401 Unauthorized
www-authenticate: Bearer {token}
Content-Length: 0
```

**To Reproduce**
Call endpoint that requires valid JWT token.

Current response:
```
HTTP/1.1 401 Unauthorized
Content-Type: application/problem+json

{
  "status": 401,
  "title": "Unauthorized",
  "instance": "/resource"
}
```

**Expected behavior**
Response should include `WWW-Authenticate` header with proper auth method (not only bearer {token}):
```
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer {token}
Content-Type: application/problem+json

{
  "status": 401,
  "title": "Unauthorized",
  "instance": "/resource"
}
```

### LINKED ISSUE #237: `HttpAuthenticator` is ignored when authentication fails
**Describe the bug**
In some setups (e.g oidc) when authentication check fails and throws `UnauthorizedException` or `AuthenticationFailedException` then exception mapper should not blindly return http 401, it should look at `HttpAuthenticator` whether there's a `ChallengeData` provided by custom auth handlers (like .oidc) and use it to build the response.

**To Reproduce**
TODO

**Expected behavior**
Classic: https://github.com/quarkusio/quarkus/blob/main/extensions/resteasy-classic/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/UnauthorizedExceptionMapper.java#L42
Reactive: https://github.com/quarkusio/quarkus/blob/main/extensions/resteasy-reactive/quarkus-resteasy-reactive/runtime/src/main/java/io/quarkus/resteasy/reactive/server/runtime/exceptionmappers/SecurityExceptionMapperUtil.java

**Ref**
Probably closes #108
https://github.com/jhipster/generator-jhipster-quarkus/issues/248#

lwitkowski:
Closes #237
Closes #108

### DIFF
--- a/deployment/src/main/java/com/tietoevry/quarkus/resteasy/problem/deployment/ProblemProcessor.java
+++ b/deployment/src/main/java/com/tietoevry/quarkus/resteasy/problem/deployment/ProblemProcessor.java
@@ -15,6 +15,7 @@ import io.quarkus.deployment.builditem.LiveReloadBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 import io.quarkus.jsonb.spi.JsonbSerializerBuildItem;
 import io.quarkus.resteasy.common.spi.ResteasyJaxrsProviderBuildItem;
+import io.quarkus.resteasy.reactive.spi.CustomExceptionMapperBuildItem;
 import io.quarkus.resteasy.reactive.spi.ExceptionMapperBuildItem;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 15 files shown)


pazkooda (approved):

lwitkowski:
Dzieki za review @pazkooda ! :)

PR #260, opened 2023-01-22
Return HTTP 500 instead of 400 on ValidationException (#254)

### LINKED ISSUE #254: ValidationException should not result in HTTP 400 status
An exception of class `javax/jakarta.validation.ValidationException` is handled by the `ValidationExceptionMapper`, which maps it to an HTTP 400 status and doesn't log the exception.

However, `ValidationException` and its subclasses (except `ConstraintViolationException`) are _not_ thrown due to failed validation, but are instead thrown on invalid use of the framework (e.g. unexpected exceptions thrown from validators).

This can clearly be seen from the names of the sub-classes:
- `ConstraintDeclarationException`
- `ConstraintDefinitionException`
- `NoProviderFoundException`
- `ValueExtractorDefinitionException`

The exception mapper in Quarkus is also quite clear about this exception [not being thrown as a result of failed validation](https://github.com/quarkusio/quarkus/blob/main/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/jaxrs/ResteasyViolationExceptionMapper.java):

> // Not a violation in a REST endpoint call, but rather in an internal component.
> // This is an internal error: handle through the QuarkusErrorHandler,
> // which will return HTTP status 500 and log the exception.

I believe it would be more correct for these exceptions to result in an HTTP 500 status being returned, and the exception logged.
I'll be happy to create a pull request that fixes this.

Tested with `com.tietoevry.quarkus:quarkus-resteasy-problem:2.1.0` using Quarkus 2.14.3.

jburkal:
Fixes #254 by having `ValidationExceptionMapper` return a HttpProblem with `INTERNAL_SERVER_ERROR` instead of `BAD_REQUEST`.

I tried deleting the `ValidationExceptionMapper` in the hope that the exception would then be handled by the `DefaultExceptionMapper`, but that didn't work as it was instead handled by the exception mapper from Quarkus.

Also adds an additional integration-test, checking that a HTTP 500 is returned if a subclass of `ValidationException` is thrown, and renames `ValidationExceptionsResource` to fix a typo.

The branch passes both `run-jvm-tests` and `run-native-tests`, but fails when running `./mvnw verify` with an exception:
```
java.lang.IllegalStateException: The current thread cannot be blocked: vert.x-eventloop-thread-0
```
However, as the `master` branch fails with the same exception on my machine, I assume it's unrelated to the changes made in this branch.

### DIFF
--- a/integration-test/src/main/java/com/tietoevry/quarkus/resteasy/problem/ValidationExceptionsResource.java
+++ b/integration-test/src/main/java/com/tietoevry/quarkus/resteasy/problem/ValidationExceptionsResource.java
@@ -2,6 +2,7 @@ package com.tietoevry.quarkus.resteasy.problem;
 
 import javax.validation.Valid;
 import javax.validation.ValidationException;
+import javax.validation.constraints.AssertFalse;
 import javax.validation.constraints.Min;
 import javax.ws.rs.GET;
 import javax.ws.rs.HeaderParam;
@@ -15,12 +16,17 @@ import org.hibernate.validator.constraints.Length;
 
 @Path("/throw/validation/")
 @Produces(MediaType.APPLICATION_JSON)
-public class ValdationExceptionsResource {
+public class ValidationExceptionsResource {
 
     @GET
     @Path("/violation-exception")
     public void throwViolationException(@QueryParam("message") String message) {
         throw new ValidationException(message);
+    }
+
+    @GET
+    @Path("/constraint-declaration-exception")
+    public void throwConstraintDeclarationException(@AssertFalse @QueryParam("message") String message) {
     }
 
     @POST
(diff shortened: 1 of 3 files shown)


jburkal:
After merging this, the documentation at https://github.com/tietoevry/quarkus-resteasy-problem/wiki#javaxvalidationvalidationexception should be updated.

lwitkowski:
> I tried deleting the `ValidationExceptionMapper` in the hope that the exception would then be handled by the `DefaultExceptionMapper`, but that didn't work as it was instead handled by the exception mapper from Quarkus.

Yeap, exceptions (re)thrown from Exception Mappers are not routed through other mappers...
 
> Also adds an additional integration-test, checking that a HTTP 500 is returned if a subclass of `ValidationException` is thrown, and renames `ValidationExceptionsResource` to fix a typo.

Excellent.

> The branch passes both `run-jvm-tests` and `run-native-tests`, but fails when running `./mvnw verify` with an exception:
> 
> ```
> java.lang.IllegalStateException: The current thread cannot be blocked: vert.x-eventloop-thread-0
> ```
> 
> However, as the `master` branch fails with the same exception on my machine, I assume it's unrelated to the changes made in this branch.

Don't worry about it then, I'll look into it separately, maybe there are some defaults for maven profiles missing.

PR #278, opened 2023-03-24
Quarkus 3.0.0.X

lwitkowski:


pazkooda:
And what about
`!Resteast jackson quickstart doesn't work yet in 3.0!`

in README.md?

Anyway I can help get it working?

lwitkowski:
There has been some changes in quickstarts, it's possible that some are completely missing in 3.0 - it needs to be investigated deeper.
This is not a blocker for releasing v3.0.0 though, as this test only validates if examples in README are still functioning.

We still have around 2 weeks till quarkus-universe-bom 3.0.0.Final is released. Quarkus core 3.0.0.Final has just been released, and now it's time for extensions (from quarkiverse and external ones) to catch up.

PR #313, opened 2023-10-27
ConstraintViolationException HTTP status should be configurable

### LINKED ISSUE #273: ConstraintViolationException HTTP status should be configurable
Hello,

some people and companies favor the usage of HTTP-Status Code 422 Unprocessable Entity over 400 Bad Request for validation errors.

I believe it would be convient, if devs could configure the Response Status Code for ConstraintViolationExceptions by a property.

lwitkowski:
Turns out `HttpProblem` relied on `Response.Status` a bit too much, which is being addressed in #314 

Closes #273

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,9 +181,8 @@ You may also want to check [this article](https://dzone.com/articles/when-http-s
 More on throwing problems: [zalando/problem usage](https://github.com/zalando/problem#usage)
 
 ## Configuration options
-All configuration options are build-time properties, meaning that you cannot override them in the runtime (i.e via environment variables).
 
-- Include MDC properties in the API response (you have to provide those properties to MDC using `MDC.put`)
+- (Build time) Include MDC properties in the API response. You have to provide those properties to MDC using `MDC.put`
 ```
 quarkus.resteasy.problem.include-mdc-properties=uuid,application,version
 ```
@@ -198,7 +197,24 @@ Result:
 }
 ```
 
-- Enable Smallrye (Microprofile) metrics for http error counters. Requires `quarkus-smallrye-metrics` in the classpath.
+- (Runtime) Changes default `400 Bad request` response status when `ConstraintViolationException` is thrown (e.g. by Hibernate Validator)
+```
+quarkus.resteasy.problem.constraint-violation.status=422
+quarkus.resteasy.problem.constraint-violation.title=Constraint violation
+```
+Result:
+```json
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/problem+json
+
+{
+    "status": 400,
+    "title": "Constraint violation",
+    (...)
+}
+```
+
+- (Build time) Enable Smallrye (Microprofile) metrics for http error counters. Requires `quarkus-smallrye-metrics` in the classpath.
 
 Please note that if you use `quarkus-micrometer-registry-prometheus` you don't need this feature - http error metrics will be produced regardless of this setting or presence of this extension.
 
@@ -212,7 +228,7 @@ application_http_error_total{status="401"} 3.0
 application_http_error_total{status="500"} 5.0
 ```
 
-- Tuning logging
+- (Runtime) Tuning logging
 ```
 quarkus.log.category.http-problem.level=INFO # default: all problems are logged
 quarkus.log.category.http-problem.level=ERROR # only HTTP 5XX problems are logged
--- a/README.md
+++ b/README.md
@@ -208,7 +208,7 @@ HTTP/1.1 422 Unprocessable Entity
 Content-Type: application/problem+json
 
 {
-    "status": 400,
+    "status": 422,
     "title": "Constraint violation",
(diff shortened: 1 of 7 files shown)


lwitkowski:
@pazkooda when you have time, you could have a look at this one after #314

pazkooda:
Yup, but it will be later today (evening/night) at the earliest.

README.md
```diff
@@ -198,7 +197,24 @@ Result:
...
+quarkus.resteasy.problem.constraint-violation.status=422
+quarkus.resteasy.problem.constraint-violation.title=Constraint violation
+```
+Result:
+```json
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/problem+json
+
+{
+    "status": 400,
```
lwitkowski, line 211:
```suggestion
    "status": 422,
```

pazkooda (approved):
LGTM

README.md
```diff
@@ -181,9 +181,8 @@ You may also want to check [this article](https://dzone.com/articles/when-http-s
 More on throwing problems: [zalando/problem usage](https://github.com/zalando/problem#usage)
 
 ## Configuration options
-All configuration options are build-time properties, meaning that you cannot override them in the runtime (i.e via environment variables).
 
-- Include MDC properties in the API response (you have to provide those properties to MDC using `MDC.put`)
+- (Build time) Include MDC properties in the API response. You have to provide those properties to MDC using `MDC.put`
```
chberger, line 185: Maybe it's worth to mention that this extension uses ``org.slf4j`` instead of ``jboss logging``. Thus a user has to pick the appropriated MDC ...

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/ProblemRuntimeConfig.java
```diff
@@ -0,0 +1,46 @@
...
+        static ConstraintViolationMapperConfig defaults() {
+            return new ConstraintViolationMapperConfig() {
+                @Override
+                public int status() {
+                    return 400;
+                }
+
+                @Override
+                public String title() {
+                    return "Bad Request";
```
chberger, line 29: Maybe it's worth to define a constant to keep the default values aligned.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
```diff
@@ -31,14 +31,20 @@
 @Priority(Priorities.USER)
```
chberger, line 31: This priority annotation is very **miss-leading** in case of  ``resteasy-reactive``, as the extension overrides the prioarty at build time: https://github.com/TietoEVRY/quarkus-resteasy-problem/blob/d253adac319b8422944159d3bfdeb7d1db91eda5/deployment/src/main/java/com/tietoevry/quarkus/resteasy/problem/deployment/ProblemProcessor.java#L115C1-L115C1

chberger (commented):
I'm leaving my comments here as well, I hope that's ok 😊

lwitkowski:
> I'm leaving my comments here as well, I hope that's ok 😊

That's more than ok, really appreciated!

README.md
lwitkowski: Good point. On the other hand, quick tests show that setting properties via `org.jboss.logging.MDC` just works, jboss probably propagates everything to slf4j (or is a service provider etc). I'll double check that in real app though.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/ProblemRuntimeConfig.java
lwitkowski: Yeah, I didn't like this defaults() method that much from the very beginning, will try to simplify this.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
lwitkowski: All I can say is - one does not simply create extension that is compatible with jackson/jsob + classic/reactive matrix ;)

All I can do here is to add clarifying comment, because the tradeoff of `the clean way` would be to have separate mappers for classic and reactive, which would look almost exactly the same.

lwitkowski: Look at [this comparison](https://github.com/TietoEVRY/quarkus-resteasy-problem/wiki) to see what we're trying to unify with this extension ;)

README.md
chberger: Ah cool, wasn't aware of that. Thanks for pointing this out.

PR #314, opened 2023-10-28
Reduce HttpProblem dependency on Response.StatusType to the minimum

lwitkowski:
This will allow some certain future features to be much easier to implement, it also reduces non-null checks as we're relying on `int` status code, not nullable `Response.StatusType`

### DIFF
--- a/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
+++ b/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
@@ -1,7 +1,5 @@
 package com.tietoevry.quarkus.resteasy.problem;
 
-import static jakarta.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
-
 import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 import java.net.URI;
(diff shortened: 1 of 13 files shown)


runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
```diff
@@ -97,8 +95,13 @@ public String getTitle() {
         return this.title;
     }
 
+    public int getStatusCode() {
+        return this.statusCode;
+    }
+
+    @Deprecated(forRemoval = true)
```
pazkooda, line 102: If you `@Deprecate` something it is best practice to document this in JavaDocs what part of new code replaces deprecated part (even if it is in same class).

I'd suggest also that we should mark this - in JavaDocs - for removal even for next minor version (`3.2.0`). 
In case of community having struggle migrating to new version we can always revert it in `3.2.1`.

On the other hand would like to get your point of view why we cannot keep it for developer convinence.

pazkooda (commented):
I'd prefer to have deprecation reason in code at least.

But beside this detail all other code look good.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/HttpProblem.java
lwitkowski: Good points. I wanted to get rid of this class from `HttpProblem entirely, but it's still heavily used in the builder by mappers. On the other hand, it will return null for 422 and other non-standard codes.

I'll remove this deprecation now, it's not that critical.

pazkooda: Well, with this note on `null` statuses I'm actually in favor of removing it in future.

But agree on not being critical for release.

PR #364, opened 2024-03-28
Fix for: Exception occurring when encoding the path containing unwise characters (i.e space)

### LINKED ISSUE #358: Exception occurring when encoding the path containing unwise characters (i.e space)
**Describe the bug**
The below exception is occurring when there is a space in the path which is not encoded, like `/api/v1/users/Some Space.`

`IllegalArgumentException: Illegal character in path at index 29: /api/v1/users/Some Space\n\tat java.base/java.net.URI.create(URI.java:906)\n\tat com.tietoevry.quarkus.resteasy.problem.postprocessing.ProblemDefaultsProvider.defaultInstance(ProblemDefaultsProvider.java:29)\n\tat com.tietoevry.quarkus.resteasy.problem.postprocessing.ProblemDefaultsProvider.apply(ProblemDefaultsProvider.java:24)\n\tat com.tietoevry.quarkus.resteasy.problem.postprocessing.PostProcessorsRegistry.applyPostProcessing(PostProcessorsRegistry.java:44)\n\tat com.tietoevry.quarkus.resteasy.problem.ExceptionMapperBase.toResponse(ExceptionMapperBase.java:25)\n\tat org.jboss.resteasy.core.ExceptionHandler.executeExceptionMapper(ExceptionHandler.java:139)\n\tat org.jboss.resteasy.core.ExceptionHandler.unwrapException(ExceptionHandler.java:183)\n\tat org.jboss.resteasy.core.ExceptionHandler.handleApplicationException(ExceptionHandler.java:100)\n\tat org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:344)`

Looking at the code it looks like the `defaultInstance` method in `ProblemDefaultsProvider` turns the path to a URI without any encoding.
`return context.path == null ? null : URI.create(context.path);
`

**To Reproduce**
Use a path which has a space in it for example `/api/v1/users/Some Space`

**Expected behavior**
An exception not to occur and the path gets encoded before it turns to a URI.

lwitkowski:
Fixes #358

### DIFF
--- a/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/postprocessing/ProblemDefaultsProvider.java
+++ b/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/postprocessing/ProblemDefaultsProvider.java
@@ -1,8 +1,12 @@
 package com.tietoevry.quarkus.resteasy.problem.postprocessing;
 
 import com.tietoevry.quarkus.resteasy.problem.HttpProblem;
+import com.tietoevry.quarkus.resteasy.problem.InstanceUtils;
+
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Replaces <code>null</code> value of <code>instance</code> with URI of currently served endpoint, i.e
@@ -22,19 +26,8 @@ final class ProblemDefaultsProvider implements ProblemPostProcessor {
         }
 
         return HttpProblem.builder(problem)
-                .withInstance(defaultInstance(context))
+                .withInstance(InstanceUtils.pathToInstance(context.path))
                 .build();
-    }
-
-    private URI defaultInstance(ProblemContext context) {
-        if (context.path == null) {
-            return null;
-        }
-        try {
-            return new URI(context.path.replaceAll(" ", "%20"));
-        } catch (URISyntaxException e) {
-            return null;
-        }
     }
 
 }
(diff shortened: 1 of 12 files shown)


runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/postprocessing/ProblemDefaultsProvider.java
```diff
@@ -26,7 +27,14 @@ public HttpProblem apply(HttpProblem problem, ProblemContext context) {
     }
 
     private URI defaultInstance(ProblemContext context) {
-        return context.path == null ? null : URI.create(context.path);
+        if (context.path == null) {
+            return null;
+        }
+        try {
+            return new URI(context.path.replaceAll(" ", "%20"));
```
pazkooda, line 34: Question out of curiosity: shouldn't we escape whole path?

pazkooda: As usual I'm referring to standards ;)

In [RFC2396](https://www.ietf.org/rfc/rfc2396.txt) (URI spec) you can find `unwise` characters that required escaping. 

```
  unwise      = "{" | "}" | "|" | "\" | "^" | "[" | "]" | "`"
```

Might not be worth doing it in this scenario though.

pazkooda (approved):
Handling of `unwise` char will be nice bonus but otherwise look OK.

runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/postprocessing/ProblemDefaultsProvider.java
lwitkowski: Thanks for details @pazkooda, very good idea to properly handle all of them (i.e by using UrlDecoder/Encoder classes), please re-review, unfortunately the changes got bigger as I needed to adjust problem-to-json serializers as well.

Issue #473, opened 2025-03-14
Incorrect Response using OpenAPI


melloware:
**Describe the bug**
When mapping using OpenApi the `HttpProblem` definition does not match the binding.

For example:

```java
@Provider
public class BundleRenderValidationExceptionMapper implements ExceptionMapper<BundleRenderValidationException> {
    @Override
    @APIResponse(
            responseCode = "400",
            description = "Validation of bundle failed",
            content = @Content(mediaType = "application/problem+json", schema = @Schema(implementation = HttpProblem.class))
    )
    public Response toResponse(BundleRenderValidationException exception) {
        return exception.getResponse();
    }
}
```

When reviewing the example document in OpenAPI:

![Image](https://github.com/user-attachments/assets/4d03fe05-38d9-41a7-b436-c15f50bc4761)

This is incorrect in multiple ways.

1. `statusCode` is actually `status` when rendered by Jackson renderer.
2.  The `headers` and `parameters` get flattened.
3. It does not have desciptions and examples of `@Schema` open api fields showing a downstream user and documenting it.
4. It generates `INCORRECT` code for clients generating from OpenAPI like Typescript.

Example flat payload with MDC trace, span, parent.

```js
{
  "status": 404,
  "title": "Not Found",
  "detail": "Logo image not found",
  "instance": "/api/v1/brand/logo",
  "traceId": "185a8b49591673349c3e7f691a1e069b",
  "spanId": "6008077d63144c9b",
  "parentId": "6008077d63144c9b"
}
```

This won't bind correctly to clients using our OpenAPI Spec.

working with @tmulle we have created one that documents well but we have to hand code any MDC variables like `X-Request-Id` which is not ideal.

```java
/**
 * Represents an HTTP Problem Response according to RFC 7807. This class
 * encapsulates details about
 * an error that occurred during HTTP request processing.
 */
@Data
@Schema(description = "HTTP Problem Response according to RFC 7807")
public class HttpProblemResponse {

    /** A URI reference that identifies the problem type */
    @Schema(description = "A URI reference that identifies the problem type", example = "https://example.com/errors/validation")
    private URI type;

    /** A short, human-readable summary of the problem type */
    @Schema(description = "A short, human-readable summary of the problem type", example = "Not Found Error")
    private String title;

    /** The HTTP status code for this occurrence of the problem */
    @JsonProperty(value = "status")
    @Schema(description = "The HTTP status code for this occurrence of the problem", example = "400")
[...]

PR #478, opened 2025-04-01
OpenApi integration as optional dependency

lwitkowski:
Done:
- added OpenApi `@Schema` definitions for `HttpProblem` and Violations while keeping open-api as optional dependency
- automatically attach `HttpProblem` `@Schema` to the OpenApi's definition of every REST endpoint that defines error responses (4XX and 5XX) but does not define explicit  schema type of the @Content

TODO:
- [ ] how to tacle custom properties in `HttpProblemSchema`, which are very specific for each project? Few ideas:
   - make `HttpProblemSchema` open for extending (non-final) by app developers - that's basically done
   - automatically extend openapi schema descriptor (in runtime) for `HttpProblem` with fields defined in `quarkus.resteasy.problem.include-mdc-properties`
- [ ] update documentation
- [ ] check open api generator results
- [ ] split integration tests into separate modules with different set of dependencies (core, zalando-problem, openapi etc) - possibly in separate PR


lwitkowski:
Hey @melloware @tmulle, I could use some feedback here, especially in the part `how to tackle custom properties in HttpProblemSchema` - I don't have a perfect solution in my mind yet, and I don't fully understand how you guys decided to solve it in your fork.

melloware:
so what we do is ALL custom properties go in Context not in the root level.

```js
{
  "status": 404,
  "title": "Not Found",
  "detail": "Aircraft not found: DL3282",
  "instance": "/api/v1/aircraft/DL3282/notes",
  "context": {
    "parentId": "b7f972037bad3d4a",
    "spanId": "b7f972037bad3d4a",
    "traceId": "d0b5fe458145e16dff3fde10af15f3ab"
  }
}
```

That way in all OpenAPI clients it maps to a `Map<String, String> context`

melloware:
here is a full payload with the new `errors` concept from the new RFC.

`context` gets custom properties and errors go in `errors` bucket.

```js
{
  "status": 422,
  "title": "Unprocessable Entity",
  "instance": "/api/v1/aircraft/1/note/1",
  "context": {
    "spanId": "c6b29f161c56248e",
    "traceId": "b3ff6a27f238a5e42514e7f607cd6efc"
  },
  "errors": [
    {
      "in": "body",
      "detail": "size must be between 0 and 2"
    }
  ]
}
```

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
```diff
@@ -0,0 +1,45 @@
+package io.quarkiverse.resteasy.problem.openapi;
+
+import java.net.URI;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(name = "HttpProblem", description = "HTTP Problem Response according to RFC9457 & RFC7807")
+public class HttpProblemSchema {
```
pazkooda, line 8: BTW. Although I'm not "happy" 😉  looking at this as kind of duplicate of `HttpProblem` I do prefer this than exposing (even correctly set-up) `HttpProblem` directly aka _never leak your implementation_.

👏

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
```diff
@@ -0,0 +1,45 @@
...
+    String detail;
+
+    /**
+     * A URI reference that identifies the specific occurrence of the problem
+     */
+    @Schema(description = "A URI reference that identifies the specific occurrence of the problem", examples = "https://api.example.com/errors/123")
+    URI instance;
+
+    /** Additional parameters providing more details about the problem */
+    // TODO this needs some thinking
```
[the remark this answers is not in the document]
melloware: @pazkooda my answer is yes because I want these details to go to clients and they need to be able to deserialize them.  I describe the exact problem in this ticket: https://github.com/quarkiverse/quarkus-resteasy-problem/issues/473

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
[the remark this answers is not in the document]
pazkooda: The way you use it seems to favor `class` though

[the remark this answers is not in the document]
tmulle: Just curious, why would you want that to be an interface? Jackson needs to have that as a POJO to deserialize the information off the wire.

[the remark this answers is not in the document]
lwitkowski: @pazkooda To be frank, I was not planning to use this class for anything else than a skeleton for openapi annotations, and I also though about interface, but it did not work for some reason. Maybe I'll double check this again.

@tmulle as stated above, this is just a skeleton to generate correct open api schema definition. It's not supposed to be used for api client error handling (parsing error response body and throwing HttpProblem) - and I would keep this use case outside of this PR to not make it bigger than it has to be.

I have this at the back of my head, and my initial idea is to have low level json deserializer that will parse application/problem+json input stream into generic json structure, pull certain fields from it by name ("title", "status" and all non-standard custom properties like mdc ones) and build `HttpProblem` from scratch without the using `readValue(json, HttpProblem.class)` type of utils. A bit more work maybe, but we will not have to change `HttpProblem` class at all.

lwitkowski: Not perfect, not terrible as they say :) Maybe for v4.0.0 release we can unify this.

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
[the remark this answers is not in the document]
lwitkowski: I'm leaning towards removing it from this schema, and leave full customisation (and providing documentation of custom fields) to app developers, with one improvement that will automatically 'document' (poorly I admit) fields that are specified as MDC-included-fields (`uuid` and `application` down below):

image

IMO it's not worse than having `context` field which does not provide any further details to the clients - you still need to provide additional documentation for these fields, either by providing your own `HttpProblemSchema` definition, adding cumbersome description to whole openapi docs or creating an external document. Or am I still missing something @melloware ?

edit: I addressed the `deserialization` part in the other thread down below.

lwitkowski:
> here is a full payload with the new `errors` concept from the new RFC.
> 
> `context` gets custom properties and errors go in `errors` bucket.
> 
> ```js
> {
>   "status": 422,
>   "title": "Unprocessable Entity",
>   "instance": "/api/v1/aircraft/1/note/1",
>   "context": {
>     "spanId": "c6b29f161c56248e",
>     "traceId": "b3ff6a27f238a5e42514e7f607cd6efc"
>   },
>   "errors": [
>     {
>       "in": "body",
>       "detail": "size must be between 0 and 2"
>     }
>   ]
> }
> ```

`errors` vs `violations` is a matter of personal preferences, we could even make it configurable, and switch to `errors` for 4.0.0 if new rfc suggests it.

For the `context` - we can't really change this in a backward compatible way, so it has to stay as is (with some small improvements and/or suggestions in another comments).

melloware:
@lwitkowski i appreciate that but without these changes it will not deserialize into the client which defeats the whole purpose right?   The core requirement is whatever data the `HttpProblem` contains needs to be represented in `HttpProblemSchema` so Jackson can serialize and deserilize it properly using OpenAPI and any client such as Swift, TypeScript etc.

pazkooda:
You'd be required to provide you Extensions into separate `Schema`-like class (with documentation) and register mapper for it.

There is no simple way to satisfy all possible Extensions we can imagine.

lwitkowski:
> @lwitkowski i appreciate that but without these changes it will not deserialize into the client which defeats the whole purpose right? The core requirement is whatever data the `HttpProblem` contains needs to be represented in `HttpProblemSchema` so Jackson can serialize and deserilize it properly using OpenAPI and any client such as Swift, TypeScript etc.

No, `application/problem+json` deserialization into `HttpProblem` was not the goal of this PR, it brings a good foundation for openapi integration, already provides value (autogenerated parts of openapi specs) and fixes some issues like `typescript codegen produces invalid structures`. 

Deserialization will follow (https://github.com/quarkiverse/quarkus-resteasy-problem/pull/478#discussion_r2023046023), I don't want to inflate this PR too much.

lwitkowski:
> (...) `HttpProblem` contains needs to be represented in `HttpProblemSchema` so Jackson can serialize and deserilize it properly using OpenAPI and any client such as Swift, TypeScript etc.

`HttpProblem` and `HttpProblemSchema` are not part of rest api specification per se - generated openapi yaml (or json) is. yaml must be compatible with real responses rest api returns. typescript, swift generators know nothing about java classes, they only know openapi yaml - this is client-server contract. And this PR aims at introducing correct yaml for generic http problem defined in RFC. It does not introduce perfect http problem for all possible applications - it's impossible due to the dynamic nature this structure.

I see zero difference whether you put your custom fields into root of json, or in a map called `context` - your api client still has no idea what to expect in those fields if you don't document it properly.

melloware:
Maybe we are talking past each other but ...

```js
{
  "status": 404,
  "title": "Not Found",
  "detail": "Aircraft not found: DL3282",
  "instance": "/api/v1/aircraft/DL3282/notes",
  "parentId": "b7f972037bad3d4a",
   "spanId": "b7f972037bad3d4a",
   "traceId": "d0b5fe458145e16dff3fde10af15f3ab"
}
```

Means I need a custom OpenAPI schema that defines what `parentId`, `spanId`, `traceId` are to deserialize them in Swift/Typescript whatever because that is ALL the client will know from the OpenAPI spec.  This is exactly the issue @tmulle and i started down this road to avoid this. 

with this model....

```js
{
  "status": 404,
  "title": "Not Found",
  "detail": "Aircraft not found: DL3282",
  "instance": "/api/v1/aircraft/DL3282/notes",
  "context": {
    "parentId": "b7f972037bad3d4a",
    "spanId": "b7f972037bad3d4a",
    "traceId": "d0b5fe458145e16dff3fde10af15f3ab"
  }
}
```

it works out of the box and context can contain any values and allow the client to decide what to  use out of it but at least it deserializes into an "object" on the client side.

lwitkowski:
> Maybe we are talking past each other but ...
> (...)
> 
> Means I need a custom OpenAPI schema that defines what `parentId`, `spanId`, `traceId` are to deserialize them in Swift/Typescript whatever because that is ALL the client will know from the OpenAPI spec. This is exactly the issue @tmulle and i started down this road to avoid this.
> 
> with this model....
> (...)
> 
> it works out of the box and context can contain any values and allow the client to decide what to use out of it but at least it deserializes into an "object" on the client side.

Ok, now I **think** I understand - you're concern about code generation and schema definition for various code generators out there, which I consider yet another of course related, but separate topic, even from deserialization I was referring to up there. Separate, but important, don't get me point. And it's also on my todo list. 

`additionalProperties` in theory should solve our issue but it all depends on specific generator and it's implementation.

What would definitely help is which generators and how you use them - that would help me test different approaches.

melloware:
@lwitkowski yep I have a working use case: https://github.com/melloware/quarkus-primereact

you just write your Quarkus code and it constantly rebuilds the TypeScript code from your Rest Services and updates the React app.

the code it generates from OpenAPI on the fly is right here: https://github.com/melloware/quarkus-primereact/blob/main/src/main/webui/src/service/CarService.ts

Specifically it generates the proper HttpProblem here:
https://github.com/melloware/quarkus-primereact/blob/c24e52bda83cc316cb5fcf335b24999a2d560ff1/src/main/webui/src/service/CarService.ts#L79-L104

Then in my React client which calls the backend I can tell Axios that all errors get deserialized to an HttpProblem typescript object.
https://github.com/melloware/quarkus-primereact/blob/c24e52bda83cc316cb5fcf335b24999a2d560ff1/src/main/webui/src/service/AxiosMutator.ts#L26

Which then allows me to generically handle all HTTP issues in using proper Typscript and handle the error
https://github.com/melloware/quarkus-primereact/blob/c24e52bda83cc316cb5fcf335b24999a2d560ff1/src/main/webui/src/CrudPage.tsx#L212-L215

lwitkowski:
> @lwitkowski yep I have a working use case: https://github.com/melloware/quarkus-primereact

Thanks, that's very helpful.
 
> you just write your Quarkus code and it constantly rebuilds the TypeScript code from your Rest Services and updates the React app.
> 
> the code it generates from OpenAPI on the fly is right here: https://github.com/melloware/quarkus-primereact/blob/main/src/main/webui/src/service/CarService.ts
> 
> Specifically it generates the proper HttpProblem here: https://github.com/melloware/quarkus-primereact/blob/c24e52bda83cc316cb5fcf335b24999a2d560ff1/src/main/webui/src/service/CarService.ts#L79-L104

Can you check how `export interface HttpProblem` looks like when you set `additionalProperties: true` next to `properties` in `#/components/schemas/HttpResponse` in your openapi yaml? 
As I edited my previous comment - it works very promising for java (it adds `HashMap<String, Object>` as superclass of a generated `HttpProblem`).

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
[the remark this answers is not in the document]
lwitkowski: Decision applied: https://github.com/quarkiverse/quarkus-resteasy-problem/pull/478/commits/e3447977cc4a60a65b436560a454e752fd29086a

melloware:
that is great

melloware:
Oh I see you are asking me to try that and regenerate. Let me test it tomorrow.

lwitkowski:
> https://github.com/melloware/quarkus-primereact

This is Typescript produced by orval for `HttpProblem` with `additionalProperties: true`:
```
/**
 * HTTP Problem Response according to RFC9457 & RFC7807
 */
export interface HttpProblem {
  /** A optional URI reference that identifies the problem type */
  type?: string;
  /** A optional, short, human-readable summary of the problem type */
  title?: string;
  /** The HTTP status code for this occurrence of the problem */
  status?: number;
  /** A optional human-readable explanation specific to this occurrence of the problem */
  detail?: string;
  /** A URI reference that identifies the specific occurrence of the problem */
  instance?: string;
  [key: string]: unknown;
}
```

I'm far from being a TS expert, but looks like `[key: string]: unknown;` line makes the whole structure dynamic, so you can get value by any key from it (basically turns it into object with few defined, strongly typed fields. Or Map if you will). 

Isn't it a solution to our problem? IMO it is, that's why I wrote before:
> I see zero difference whether you put your custom fields into root of json, or in a map called context - your api client still has no idea what fields to expect there unless app developer documents it properly.

melloware:
Yes that is correct with key of string... the issue is can OpenApi be marked up with AdditionalProperties from the Quarkus Side ?  That is why I put it in "context" so it would bind to a map in a more strongly typed location but I am ok if your OpenApi generates it as additional properties. 

Now comes the problems with violations and Cause 🤣

lwitkowski:
Yeah, you can tell Quarkus to set this flag to for any type you want, look in this PR for '@Schema' on 'HttpProblemSchema' class, it's pretty self explanatory.

For constraint violation I intruduced separate schema class extending base 'HttpProblemSchema', it serves the purpose at first glance.
What issues do you have in mind specifically?

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
```diff
@@ -0,0 +1,39 @@
+package io.quarkiverse.resteasy.problem.openapi;
+
+import java.net.URI;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(name = "HttpProblem", description = "HTTP Problem Response according to RFC9457 & RFC7807", additionalProperties = Schema.True.class)
+public class HttpProblemSchema {
```
melloware, line 8: Maybe this class should be `HttpProblemResponse` instead of `Schema`?   Because on the other end it should decode to a domain object and the name `HttpProblemSchema` does not feel like a domain object but `HttpProblemResponse` describes what it actually is?

melloware (changes requested):

melloware:
So for `cause` we have this: https://github.com/melloware/quarkus-primereact/blob/c24e52bda83cc316cb5fcf335b24999a2d560ff1/src/main/webui/src/service/CarService.ts#L103

this was because when using MicroProfile Client @tmulle needs to get the original cause and forward it on with a new problem when using multiple Microservices and Quarkus MicroProfile Client.

melloware:
@tmulle even wrote some sepearate integration tests testing the client: https://github.com/melloware/quarkus-openapi-problem/tree/main/integration-tests-client

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
melloware: Nevermind I see you override the name `name = "HttpProblem"`. Please ignore.

melloware:
I am going to pull this branch today and build and test my scenarios and let you know how it goes!

runtime/src/main/java/io/quarkiverse/resteasy/problem/openapi/HttpProblemSchema.java
lwitkowski: TIL (well, today I understood, I learned few days ago) that JVM class loader does not throw `ClassNotFoundExceptions` for annotations that are not present in the classpath, and this is as-designed behaviour O_o
https://stackoverflow.com/questions/3567413/why-doesnt-a-missing-annotation-cause-a-classnotfoundexception-at-runtime

This means we can add these annotations directly on the main `HttpProblem` class without breaking backward compatibility (and without adding non-optional dependency). I'm gonna test it before this PR is ready for review, and maybe HttpProblem and HttpProblemSchema become one.

lwitkowski: @pazkooda I've just pushed this change, now `HttpProblem` itself is annotated with openapi stuff so duplication is kind of gone. I'm not sure if I understood your original comment about exposing `HttpProblem` directly so not sure how/if you'll like this change/proposal. In the end `HttpProblem` is already exposed 'everywhere' and is part of public programmatic api of this extension.

melloware:
This is great....plan on getting to this today

pazkooda:
Will take a look / review morning tomorrow at the latest.

Great work @lwitkowski .

melloware:
Just tested my use case and everything looks good!  @tmulle you should test this as well.

melloware (approved):

runtime/src/main/java/io/quarkiverse/resteasy/problem/ProblemRuntimeConfig.java
```diff
@@ -6,8 +6,12 @@
 import io.smallrye.config.WithDefault;
 import io.smallrye.config.WithName;
 
+/**
+ * @deprecated v4.0.0 all configuration will be evaluated in build time
```
pazkooda, line 10: Consider suplementing this with link to new version of _config_, sth like:
```java
/**
* ...
* <p> Use {@link ProblemBuildConfig} instead.
*/
```

runtime/src/main/java/io/quarkiverse/resteasy/problem/security/ForbiddenExceptionMapper.java
```diff
@@ -13,7 +16,10 @@
  * Mapper overriding default Quarkus exception mapper to make all error responses compliant with RFC7807.
  */
 @Priority(Priorities.USER)
-public final class ForbiddenExceptionMapper extends ExceptionMapperBase<ForbiddenException> {
+// OpenApi annotation is actually a build time dependency, not runtime!
```
pazkooda, line 19: For clean-up I assume

runtime/src/main/java/io/quarkiverse/resteasy/problem/validation/HttpValidationProblem.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Representation of RFC7807 Problem schema.
+ */
+@Schema(name = "HttpValidationProblem", description = "HTTP Validation Problem Response according to RFC9457 & RFC7807", additionalProperties = Schema.True.class)
+public class HttpValidationProblem extends HttpProblem {
+
+    @Schema(description = "List of validation constraint violations that occurred")
+    List<Violation> violations;
+
+    public HttpValidationProblem(List<Violation> violations, int status, String title) {
```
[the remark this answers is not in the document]
lwitkowski: Agree, fixed.

runtime/src/main/java/io/quarkiverse/resteasy/problem/security/ForbiddenExceptionMapper.java
lwitkowski: Removed.

runtime/src/main/java/io/quarkiverse/resteasy/problem/ProblemRuntimeConfig.java
lwitkowski: Link added

pazkooda:
> ...
>     * default openapi schema for problem responses is now configurable, which allows developers to provide their own 'gold standard' of Problem response schema:
> 
> 
> ```java
> @Schema(name = "MyHttpProblem", description = "HTTP Problem Response according to MyProject",
>         additionalProperties = Schema.True.class)
> public class MyHttpProblem extends HttpProblem {
> 
>     @Schema(description = "Additional parameters providing more details about the problem", examples = "{\"timestamp\":\"2024-03-20T10:00:00Z\",\"traceId\":\"550e8400-e29b-41d4-a716-446655440000\"}")
>     private SortedMap<String, Object> contexts;
> 
>     @Schema(description = "Original cause of error, only set when forwarding an underlying problem")
>     private MyHttpProblem cause;
> 
> }
> ```
> 
> plus this in `application.properties`:
> 
> ```ini
> quarkus.resteasy.problem.openapi.default-schema=MyHttpProblem
> ```
> 
> I did not find a way to replace existing schema with another one on openapi schemas list, so this is the best I can think of atm.
> 
>     * `HttpValidationProblem` openapi's statuscode and descriptions is taken from configuration
> 
> 
> ```ini
> quarkus.resteasy.problem.constraint-violation.status=422
> quarkus.resteasy.problem.constraint-violation.title=Unprocessable Content
> quarkus.resteasy.problem.constraint-violation.description=Unprocessable Content: server understood the content type of the request content, and the syntax of the request content was correct, but it was unable to process the contained instructions.
> ```
> ...

☝️ I assume you gonna extract this into `README.md` docs.

lwitkowski:
> > (...)
> 
> ☝️ I assume you gonna extract this into `README.md` docs.

yeap, will do that once the api and the code part is done.

lwitkowski:
Some docs added: https://github.com/quarkiverse/quarkus-resteasy-problem/tree/feat/openapi?tab=readme-ov-file#openapi-integration-available-since-v3200

The work on openapi integration is not yet done (e.g. #482), but will be addressed in separate PRs before 3.20 release (I'll try to publish SNAPSHOT version for testing after this gets merged).

Thanks @pazkooda @melloware @tmulle for all the feedback.

melloware:
@lwitkowski this is a huge improvement for this library thank you for taking the time to make it great!

PR #483, opened 2025-04-03
#482 Support for application/problem+json deserialization for RestClients

### LINKED ISSUE #429: MAJOR: Response is truncated when calling from one service to another and error happens on backend
I ran into an issue when developing a project for our company set to go live in a few days. So I'm hoping this is a simple issue.

I'm seeing the JSON output from resteasy-problem being truncated in the scenario when I make a rest call from one service to another service and that service throws an error (404 in this case).

For example, I have a reproducer at: https://github.com/tmulle/quarkus-resteasy-problem-test

I have a simulated `gateway` and `server` service in their respective projects.

To Test:

1. Check out project
2. Start both the `server` and `gateway` projects in separate terminals..I started my server on port 12000 and gateway on 12001. I hardcoded the backend in the `RemoteService` class in the gateway project. If you change ports you'll need to modify the code.
3. Send a `POST` using curl or `postman/insomnium` to http://localhost:12000/hello/notFound and observe the correct JSON error  response as shown below. This is hitting the backend `server` service directly.
4. Send a `POST` using curl or `postman/insomnium` to http://localhost:12001/gateway/notFound and observe you get the truncated JSON error message as shown below. This is hitting the `gateway` which then calls the `server` and returns the response.

Correct Response - When direct call to the backend service
```json
{
	"status": 404,
	"title": "Not Found",
	"detail": "HTTP 404 Not Found",
	"instance": "/hello/notFound"
}
```
Truncated Response - When called through gateway (the output is missing the rest of the structure)
```json
{
	"status": 404,
	"title": "Not Found",
	"detail": "Received: 'Not Found, status code 404' when invoki
```

I don't get any other errors so I'm not sure why the json is being truncated. I found this issue because my Javascript web frontend I'm developing which is calling my GATEWAY->REMOTE_SERVICE threw an exception when parsing the JSON error from my Quarkus servers which then lead me to track down the issue to this library or something near it.

If you need anymore info, let me know..

### LINKED ISSUE #482: Support for `application/problem+json` deserialization for RestClients
TODO

TODO

lwitkowski:
### Changes
- [x] jackson deserializer for `HttpProblem`
- [x] jsonb deserializer for `HttpProblem`
- [x] `ThrowingHttpProblemClientExceptionMapper` for RestClients to deserialize `application/problem+json` and throw as `HttpProblem` 
- [x] fix for #429  
- [x] documentation

Closes #482, closes #429, related to #473

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemModuleRegistrar.java
+++ b/runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemModuleRegistrar.java
@@ -13,15 +13,10 @@ public final class JacksonProblemModuleRegistrar implements ObjectMapperCustomiz
 
     @Override
     public void customize(ObjectMapper mapper) {
-        registerProblemModule(mapper);
-    }
-
-    public static ObjectMapper registerProblemModule(ObjectMapper mapper) {
-        mapper.registerModule(
-                new SimpleModule("RFC7807 problem")
-                        .addSerializer(HttpProblem.class, new JacksonProblemSerializer())
-                        .addDeserializer(HttpProblem.class, new JacksonProblemDeserializer(mapper)));
-        return mapper;
+        SimpleModule module = new SimpleModule("RFC7807 problem")
+                .addSerializer(HttpProblem.class, new JacksonProblemSerializer())
+                .addDeserializer(HttpProblem.class, new JacksonProblemDeserializer(mapper));
+        mapper.registerModule(module);
     }
 
 }
--- a/runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemModuleRegistrar.java
+++ b/runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemModuleRegistrar.java
@@ -13,9 +13,15 @@ public final class JacksonProblemModuleRegistrar implements ObjectMapperCustomiz
 
     @Override
     public void customize(ObjectMapper mapper) {
-        SimpleModule module = new SimpleModule("RFC7807 problem");
-        module.addSerializer(HttpProblem.class, new JacksonProblemSerializer());
-        mapper.registerModule(module);
+        registerProblemModule(mapper);
+    }
+
+    public static ObjectMapper registerProblemModule(ObjectMapper mapper) {
+        mapper.registerModule(
+                new SimpleModule("RFC7807 problem")
+                        .addSerializer(HttpProblem.class, new JacksonProblemSerializer())
+                        .addDeserializer(HttpProblem.class, new JacksonProblemDeserializer(mapper)));
+        return mapper;
     }
 
 }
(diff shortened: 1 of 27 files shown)


lwitkowski:
In the fix for #429 I still decided to pass all headers except the one that caused issue ('content-length'), but the more I think about it the more I believe this extension should not pass any header by default (mainly for safety reasons, but it may be also error-prone). 

Maybe we should introduce `quarkus.resteasy.problem.proxy-pass-header` config option (default empty) with explicit list of headers developers decide they want to pass from upstream service to gateway clients. In both scenarios (with `ThrowingHttpProblemClientExceptionMapper` and without).

WDYT?

runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemModuleRegistrar.java
```diff
@@ -13,9 +13,15 @@ public final class JacksonProblemModuleRegistrar implements ObjectMapperCustomiz
 
     @Override
     public void customize(ObjectMapper mapper) {
-        SimpleModule module = new SimpleModule("RFC7807 problem");
-        module.addSerializer(HttpProblem.class, new JacksonProblemSerializer());
-        mapper.registerModule(module);
+        registerProblemModule(mapper);
+    }
+
+    public static ObjectMapper registerProblemModule(ObjectMapper mapper) {
```
pazkooda, line 19: `public` for tests only 😜 

You made duplication on `uriOrThrow`/`intOrThrow` yet only for sake of single use in test you "expose API 😉 " in Quarkus Bean.

My preference is to duplicate it in test.

But since this PullReq is "so good" this is the only thing I can catch really 😄 

Good job❕

pazkooda (approved):
Beside one minor thing look really good!

lwitkowski:
> Beside one minor thing look really good!

Thanks man, would you mind having a look at this comment? https://github.com/quarkiverse/quarkus-resteasy-problem/pull/483#issuecomment-2778311060

runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemModuleRegistrar.java
lwitkowski: Agree, I took it a bit too far. Reverted this.

runtime/src/test/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemDeserializerTest.java
```diff
@@ -0,0 +1,137 @@
...
+import com.fasterxml.jackson.databind.module.SimpleModule;
+
+import io.quarkiverse.resteasy.problem.HttpProblem;
+import io.quarkiverse.resteasy.problem.HttpProblemMother;
+
+class JacksonProblemDeserializerTest {
+
+    ObjectMapper mapper = new ObjectMapper();
+
+    @BeforeEach
```
pazkooda, line 26: `@BeforeAll` might be enough enough...

melloware (approved):
This is great!

melloware:
I have archived my Repo and left a note saying that all features have been ported into this extension: 
https://github.com/melloware/quarkus-openapi-problem

image

runtime/src/test/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemDeserializerTest.java
lwitkowski: Then both `mapper` and `setup` method would have to be static

PR #491, opened 2025-04-19
Fix #489: HttpProblem doc change & to and

### LINKED ISSUE #489: Description of HttpProblem causing JavaDoc error
The schema description containing the `&` causes issues when the javadoc plugin runs because it gives an invalid entity error.

Description that is in the schema tag of HttpProblem.

```
HTTP Problem Response according to RFC9457 & RFC7807
```

Should probably escape that or just use the word `and`

melloware:
Fix #489: HttpProblem doc change & to and


pazkooda:
Thanks for contributing 👍🏻.

Please adjust test, see build failure.

melloware:
Tests resolved

pazkooda (approved):

Issue #295, opened 2023-09-07
Issue regarding propertynames in constraintviolation.


bdunni:
When I throw a plain constraing violation, violating as an example: email address.
I recieve the message in the issue, but I do not get the fieldname or what the actual value is.

I reviewed the code and found the below implementation quite weird.

Why do we visit the iterator twice to get the propertyname?

https://github.com/TietoEVRY/quarkus-resteasy-problem/blob/da642a0aad2d5892fc1ecc466edf4005c468e8db/runtime/src/main/java/com/tietoevry/quarkus/resteasy/problem/validation/ConstraintViolationExceptionMapper.java#L64-L78

lwitkowski:
Hi @bdunni, it may seem weird, but hibernate validator puts controller's class name and method name as first 2 elements of the property path - we decided we don't want to leak this implementation detail outside of REST api - api clients don't care if it's java or php, if it's handled by ResourceA, or ResourceB.

This may interfer with your use case - would you mind posting reproducer/example here?

bdunni:
Hey @lwitkowski thanks for the quick reply.

I will see if I can make a reproducer example.

It also has me questioning line 75.
We compare the parameter name in the resource to the parameter name in violation.
Am I wrong to assume that this is also interfering with my use case?

In my case I have validation on "Email" it is part of the body of my request, but all my resourceInfo method parameters delivers is the name of the DTO which email is a part of. Hence this will never be a valid?


as an example:

public class PersonDTO  {
@NotNull
String email
}

Then the parameters returned by line 75 will return PersonDTO, but never email as it is inside the object?

bdunni:
@lwitkowski wait a second. Hibernate validator?
In this scenario we are validating on the incoming request and not on entities. Am I perhaps throwing Constraintviolations wrongly?

NissenDev:
@lwitkowski I have to post from another account.

I reproduced the problem here:
https://github.com/NissenDev/Http-problem-reproduce

Simply make a post with the following Json:
{
    "email": 222222,
    "title": "45654565"
}

To http://localhost:8080/hello

I used POSTMAN

bdunni:
So it seems to be that we do not support the @RequestBody as a param of the ResourceInfo methods.
Furthermore the 2 first elements of the controller and name and class are not present when doing the post.

lwitkowski:
[...]

PR #510, opened 2025-07-04
#295 Programmatic validation property path handling 

pazkooda:
@lwitkowski Hey,

I've asked Claude Code to try to address #295 and it came with this solution.
I know it's bit dirty (commits, maybe a bit code structure)

But this Draft PullReq is to discuss if we are OK with the concept itself.

I can do clean-up later on.

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
+++ b/runtime/src/main/java/io/quarkiverse/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
@@ -68,11 +68,23 @@ public final class ConstraintViolationExceptionMapper extends ExceptionMapperBas
     }
 
     private Violation toViolation(ConstraintViolation<?> constraintViolation) {
-        return matchEndpointMethodParameter(constraintViolation)
-                .map(param -> createViolation(constraintViolation, param))
-                .orElseGet(() -> Violation.In.unknown
-                        .field(dropMethodName(constraintViolation.getPropertyPath()))
-                        .message(constraintViolation.getMessage()));
+        // First try to match JAX-RS parameter (declarative case)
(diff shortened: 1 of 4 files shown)


pazkooda:
@bdunni Curious if you could build this branch locally and validate if it works in the case from original issue.

If you could provide a test with validation groups this would be perfect.

lwitkowski:
@pazkooda Nice! Looks pretty good already, some comments seems a bit redundant, small code cleanups here and there, and we should merge this.

pazkooda:
@lwitkowski My vacation ended up being extented till yesterday 👉  delay in making this review-able.

Please take a look and see if you see possible improvments to be introduced.

runtime/src/main/java/io/quarkiverse/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
```diff
@@ -140,4 +147,24 @@ private String serializePath(Path propertyPath, int skipFirstSegments) {
         return String.join(".", pathSegments);
     }
 
+    private boolean isDeclarativeValidation(ConstraintViolation<?> violation) {
+        return hasJaxRsContext() &&
+                (rootBeanResourceClassMatches(violation) || propertyPathStartsWithMethod(violation));
```
lwitkowski, line 152: Nitpick: I personally find it more readable to expand complex if conditions into multiple ifs like this:
```suggestion
        if(!hasJaxRsContext()) {
            return false;
        }
        if(rootBeanResourceClassMatches(violation)) {
            return true;
        }
        if(propertyPathStartsWithMethod(violation))) {
            return true;
        }
        return false;
```

But feel free to ignore this suggestion.

lwitkowski (approved):

PR #520, opened 2025-08-04
fix #524: trailing dot in constraint violation paths

### LINKED ISSUE #524: Trailing dot in validation error message.
### Discussed in https://github.com/quarkiverse/quarkus-resteasy-problem/discussions/519

<div type='discussions-op-text'>

<sup>Originally posted by **SawantKunal** August  4, 2025</sup>
I'm using `quarkus-resteasy-problem` with custom validations. When I'm not using `addPropertyNode` method, I noticed that when validation fails, the constraint violation path includes an unnecessary trailing dot. 

Custom validation code -
``` java
@Override
  public boolean isValid(Code code, ConstraintValidatorContext constraintValidatorContext) {
    boolean isValid = true;
    constraintValidatorContext.disableDefaultConstraintViolation();

    if (pattern != null && !pattern.matcher(code.code()).matches()) {
      constraintValidatorContext.buildConstraintViolationWithTemplate("must match \"" + pattern.pattern() + "\"")
          .addConstraintViolation();
      isValid = false;
    }
    return isValid;
  }
```
Example validation error message -
``` json
"violations": [
    {
      "field": "ifsc.",
      "in": "body",
      "message": "must match regex"
    }
]
```
This seems to be caused by empty path segments being included during path serialization. I’ve fixed the issue by adding a null/empty check when serializing the Path, and I’m raising a PR to propose the change. Would appreciate feedback!</div>

SawantKunal:
As per this [discussion](https://github.com/quarkiverse/quarkus-resteasy-problem/discussions/519) I have raise the PR. I have updated ConstraintViolationExceptionMapper to skip null or empty path segments when building the property path. This was added to avoid trailing dot in constraint violation paths for value objects.

### DIFF
--- a/integration-test/core/src/main/java/io/quarkiverse/resteasy/problem/ValidationExceptionsResource.java
+++ b/integration-test/core/src/main/java/io/quarkiverse/resteasy/problem/ValidationExceptionsResource.java
@@ -1,10 +1,19 @@
 package io.quarkiverse.resteasy.problem;
 
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import jakarta.inject.Inject;
+import jakarta.validation.Constraint;
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
 import jakarta.validation.ConstraintViolation;
 import jakarta.validation.ConstraintViolationException;
+import jakarta.validation.Payload;
 import jakarta.validation.Valid;
 import jakarta.validation.ValidationException;
 import jakarta.validation.Validator;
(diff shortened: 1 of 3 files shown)


runtime/src/main/java/io/quarkiverse/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
```diff
@@ -141,7 +141,10 @@ private String serializePath(Path propertyPath, int skipFirstSegments) {
                 skipFirstSegments -= 1;
                 segmentIterator.next();
             } else {
-                pathSegments.add(segmentIterator.next().toString());
+                String segment = segmentIterator.next().toString();
+                if (segment != null && !segment.trim().isEmpty()) {
+                    pathSegments.add(segment);
+                }
```
lwitkowski, line 147: I think it's a bit too late to check for segment nullability, IMO it never can be null, because `segmentIterator.next()` never returns null thanks to `hasNext()` check above.

also, `segment.trim().isEmpty()` has shorter version: `segment.isBlank()` since java 11.

```suggestion
                String segment = segmentIterator.next().toString();
                if (!segment.isBlank()) {
                    pathSegments.add(segment);
                }
```

lwitkowski (changes requested):
Tests look good, few suggestions in the comments, please have a look.

integration-test/core/src/main/java/io/quarkiverse/resteasy/problem/ValidationExceptionsResource.java
```diff
@@ -121,4 +127,44 @@ public static final class ProgrammaticTestAddress {
...
+      @Valid
+      public CustomName name;
+    }
+
+    @ValidCustomName
+    public static final class CustomName {
+      public String code;
+    }
+
+    @jakarta.validation.Constraint(validatedBy = CustomNameValidator.class)
```
pazkooda, line 141: Just curious: looks like generated by LLM tool.
Can you move this package to `import` and leave only `Constraint` here?
Incl. code below.

integration-test/core/src/main/java/io/quarkiverse/resteasy/problem/ValidationExceptionsResource.java
```diff
@@ -121,4 +127,44 @@ public static final class ProgrammaticTestAddress {
...
+    @java.lang.annotation.Target({ java.lang.annotation.ElementType.TYPE })
+    @java.lang.annotation.Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
```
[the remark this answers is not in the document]
pazkooda: My guess: LLM 😜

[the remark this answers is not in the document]
pazkooda: So it is LLM, but genereted by me 😅 .

I've asked Claude Code to generate a test draft. It failed to pinpoint problem so I've deleted the comment with draft.

It's not a problem. Let's just clean it up as we really appreciated contribution (with tests included)

integration-test/core/src/main/java/io/quarkiverse/resteasy/problem/ValidationExceptionsResource.java
```diff
@@ -121,4 +136,43 @@ public static final class ProgrammaticTestAddress {
...
+      public boolean isValid(CustomName customName, ConstraintValidatorContext constraintValidatorContext) {
+        String code = customName.code;
+        constraintValidatorContext.disableDefaultConstraintViolation();
+
+        if (pattern != null && !pattern.matcher(code).matches()) {
+          constraintValidatorContext.buildConstraintViolationWithTemplate("must match \"" + pattern.pattern() + "\"")
+              .addConstraintViolation();
+          return false;
+        }
+        return true;
```
lwitkowski, line 174: You missed part of my comment/suggestion:
`pattern` is final now, so it's never null - check is redundant. Also, please either inline `code` variable, or move it right above the place where it is used for the first time. Right now the sections that do different things are tangled - first you prepare tmp variable, then call some modifier method on the context, then you use this variable later.

This should do:
```suggestion
        constraintValidatorContext.disableDefaultConstraintViolation();

        String code = customName.code;
        if (!pattern.matcher(customName.code).matches()) {
          constraintValidatorContext.buildConstraintViolationWithTemplate("must match \"" + pattern.pattern() + "\"")
              .addConstraintViolation();
          return false;
        }
        return true;
```

PR #526, opened 2025-08-12
Fix for #517: JacksonProblemDeserializer does not support JsonInclude.Include.ALWAYS

### LINKED ISSUE #517: JacksonProblemDeserializer does not support JsonInclude.Include.ALWAYS
**Describe the bug**
If the rest-client you are calling uses JsonInclude.Include.ALWAYS so that it for example responds with:
```json
{
   "status": 401,
   "detail": null,
   "title": "The JWT has expired",
    "type": null,
    "instance": "test/get"
 }
```
Then you will get a null pointer exception since uriOrThrow() will still be called since the key is present. Same is true for **instance**

**To Reproduce**
This code reproduces the error:
```
public class DemoResource {

    @Inject
    ObjectMapper objectMapper;
 
    @GET
    @Path("/test-object-mapper")
    public void objectMapper() {
        String json = """
            {
              "status": 401,
              "detail": null,
              "title": "The JWT has expired",
              "type": null,
              "instance": "test/get"
            }
            """;
        try {
            objectMapper.readValue(json, HttpProblem.class);
        }catch (JsonProcessingException e) {
            throw new RuntimeException("Failed to deserialize JSON to HttpProblem", e);
        }
    }
}
```
```
@QuarkusTest
class RestClientIT {

    static {
        RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
    }

    @Test
    void objectMapper() {
        given()
            .accept(ContentType.JSON)
            .get("/test-object-mapper")
            .then()
            .statusCode(500);
    }

}
```

**Expected behavior**
I expect the object mapper to skip these properties when their key is present but the value is null aka something like this:
```
            Object child = rawDeserializedProblem.get(fieldName);
            if (child == null) {
                // Ignore null values
                continue;
            }

            switch (fieldName) {
                case "type" -> builder.withType(uriOrThrow(child, fieldName, jsonParser));
                case "status" -> builder.withStatus(intOrThrow(child, fieldName, jsonParser));
                case "title" -> builder.withTitle((String) child);
                case "detail" -> builder.withDetail((String) child);
                case "instance" -> builder.withInstance(uriOrThrow(child, fieldName, jsonParser));
                default -> builder.with(fieldName, child);
            }
```

lwitkowski:
Closes #517

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemDeserializer.java
+++ b/runtime/src/main/java/io/quarkiverse/resteasy/problem/jackson/JacksonProblemDeserializer.java
@@ -46,6 +46,10 @@ public final class JacksonProblemDeserializer extends StdDeserializer<HttpProble
     }
 
     private URI uriOrThrow(Object child, String fieldName, JsonParser jsonParser) throws JsonMappingException {
+        if(child == null) {
+            return null;
+        }
+
         try {
             return URI.create((String) child);
         } catch (IllegalArgumentException e) {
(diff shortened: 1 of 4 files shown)


pazkooda (approved):
LTGM 👍

Issue #573, opened 2026-02-17
ConstraintViolation violation details getting lost


michaelsonnleitner:
**Describe the bug**
Version: 3.21.0

given following interface:
```
@jakarta.ws.rs.Path("/search")
@jakarta.enterprise.context.ApplicationScoped
public interface SearchResource {
    @jakarta.ws.rs.GET
    void search(
        @jakarta.ws.rs.QueryParam("query") @jakarta.validation.constraints.Size(min=10, max=128) String query);
}
```
and following implementation:
```
public class SearchService implements SearchResource {
    @Override
    public void search(String query) {
        //do it
    }
}
```

making a rest call:
`curl --location 'http://localhost:8080/search?query=short'`

produces following:
```
{
    "status": 400,
    "title": "Bad Request",
    "instance": "/search",
    "violations": [
        {
            "field": "", <-- missing
            "in": "body", <-- wrong - it is query
            "message": "Größe muss zwischen 10 und 128 sein"
        }
    ]
}
```
see debugger. Annotations of `param` in `ConstraintViolationExceptionMapper.createViolation` is gone.

<img width="793" height="190" alt="Image" src="https://github.com/user-attachments/assets/0e7f5dcb-f7da-4a91-b7ea-81f930032ac2" />

**Expected behavior**
should produce:
```
{
    "status": 400,
    "title": "Bad Request",
    "instance": "/search",
    "violations": [
        {
            "field": "query", <-- correct name
            "in": "query", <-- correct param type
            "message": "Größe muss zwischen 10 und 128 sein"
        }
    ]
}
```

lwitkowski:
[Fix released as 3.21.1](https://github.com/quarkiverse/quarkus-resteasy-problem/releases/tag/3.21.1) should now be available in Maven Central: https://repo1.maven.org/maven2/io/quarkiverse/resteasy-problem/quarkus-resteasy-problem/3.21.1/

Thanks for the report @michaelsonnleitner , thanks for the fix @pazkooda !

PR #574, opened 2026-02-17
[#573] Fix for lost param details on validation when on Interface

pazkooda:
Co-authored-by: Claude Code

@michaelsonnleitner: Can you try to build this locally, change version against your project to `SNAPSHOT` and confirm if it works.

@lwitkowski: This is done by Claude Code. But looks good enough for me to me merged.

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
+++ b/runtime/src/main/java/io/quarkiverse/resteasy/problem/validation/ConstraintViolationExceptionMapper.java
@@ -97,32 +97,32 @@ public final class ConstraintViolationExceptionMapper extends ExceptionMapperBas
     }
 
     private Optional<Parameter> findParameterInHierarchy(Method method, String paramName) {
-        Optional<Parameter> fromMethod = findParameterByName(method, paramName);
-        if (fromMethod.map(this::hasJaxRsParamAnnotation).orElse(false)) {
-            return fromMethod;
+        return Stream.concat(Stream.of(method), interfaceMethods(method))
+                .flatMap(m -> Stream.of(m.getParameters()))
+                .filter(p -> p.getName().equals(paramName))
+                .filter(this::hasJaxRsParamAnnotation)
+                .findFirst()
+                .or(() -> findParameterByName(method, paramName));
+    }
+
+    private Stream<Method> interfaceMethods(Method method) {
+        return Stream.of(method.getDeclaringClass().getInterfaces())
+                .map(iface -> findMethodInInterface(iface, method))
+                .flatMap(Optional::stream);
+    }
+
+    private Optional<Method> findMethodInInterface(Class<?> iface, Method method) {
+        try {
+            return Optional.of(iface.getMethod(method.getName(), method.getParameterTypes()));
+        } catch (NoSuchMethodException ignored) {
+            return Optional.empty();
         }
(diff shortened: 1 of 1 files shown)


pazkooda:
```markdown
Root Cause Analysis

  Where the bug lives: This extension (not Quarkus)

  The issue stems from a behavioral difference between JAX-RS implementations, but the fix belongs here.

  ---
  Root Cause

  Java reflection does not inherit parameter annotations from interfaces. This is by spec — the @Inherited meta-annotation only covers superclass
  inheritance, never interface implementation.

  When you override an interface method without re-declaring annotations:
  interface SearchResource {
      void search(@QueryParam("query") @Size(min=10, max=128) String query);
  }
  class SearchServiceImpl implements SearchResource {
      @Override
      public void search(String query) { }  // ← no annotations here
  }
  SearchServiceImpl.class.getMethod("search", String.class).getParameters()[0].getAnnotation(QueryParam.class) returns null.

  ---
  How RESTEasy Classic vs Reactive differs

  RESTEasy Classic — ResourceInfo.getResourceMethod() returns the interface method → has @QueryParam → works.

  RESTEasy Reactive — ResourceInfo.getResourceMethod() returns the implementation class method → no annotations → bug.

  This is tracked as quarkusio/quarkus#23956, still open, no resolution planned. Quarkus contributors consider the Reactive behavior (returning the
  implementation class) to be correct, so we cannot depend on Quarkus fixing this.

  ---
  Where exactly in this extension it breaks

  matchEndpointMethodParameter() (line 95–98):

  return Optional.ofNullable(resourceInfo.getResourceMethod())
          .flatMap(method -> Stream.of(method.getParameters())
                  .filter(param -> param.getName().equals(paramName))
                  .findFirst());

  It finds the parameter by name correctly, but the returned Parameter is from the implementation method (no annotations). createViolation() then sees no
  @QueryParam/@PathParam/etc. and defaults to in=body, field="".

  ---
  Possible fixes

  Option A — Walk the interface hierarchy in the mapper (recommended)

  After finding the parameter from the implementation method, check if it has JAX-RS param annotations. If not, search the same method signature across all
  implemented interfaces and return the first annotated match. This is contained entirely within the extension, requires no new dependencies, and handles
  both RESTEasy Classic (already returns the interface method → fast path) and Reactive.

  Sketch:
  // In matchEndpointMethodParameter: instead of returning the first found param...
  // → prefer interface parameter if impl param lacks JAX-RS annotations
  private Optional<Parameter> findParameterInHierarchy(Method method, String paramName) {
      Optional<Parameter> fromMethod = findParameterByName(method, paramName);
      if (fromMethod.map(this::hasJaxRsParamAnnotation).orElse(false)) {
          return fromMethod; // fast path: method already has annotations
      }
      // Fall back: walk interfaces of the declaring class
      Optional<Parameter> fromInterface = findParameterInInterfaces(method.getDeclaringClass(), method, paramName);
      return fromInterface.isPresent() ? fromInterface : fromMethod;
  }
  - Preserves existing behavior for all current tests
  - Only adds ~15 lines of private helpers
  - Handles body params correctly (no JAX-RS annotation on interface either → still falls through to body detection in createViolation)

  Option B — Search from resourceInfo.getResourceClass() interfaces

  Similar to A, but starts from resourceInfo.getResourceClass() instead of method.getDeclaringClass(). Both resolve to the same class in RESTEasy Reactive,
  but Option A is safer if the resource class hierarchy is deeper.

  Option C — Delegate to Quarkus/wait for upstream fix

  Not viable — the Quarkus issue is open and stalled, and the behavior difference is considered intentional by the Quarkus team.
  ```

PR #611, opened 2026-05-14
Fix for: NullPointerException on unauthenticated request when non-standard status code is used #601

### LINKED ISSUE #601: [Bug] NullPointerException on unauthenticated JS request with @Authenticated
**Describe the bug**
When `quarkus.oidc.authentication.java-script-auto-redirect=false` is set and a resource is annotated with `@Authenticated`, an unauthenticated request with the `X-Requested-With: JavaScript` header causes a `NullPointerException` instead of returning a proper [RFC-7807 Problem Details](https://datatracker.ietf.org/doc/html/rfc7807) response via `quarkus-resteasy-problem`.

**To Reproduce**
- Sample
  - [https://github.com/kara-g/quarkiverse-resteasy-problem-bug-repro](https://github.com/kara-g/quarkiverse-resteasy-problem-bug-repro)
- Quarkus version: 3.33.1
- quarkus-resteasy-problem version:3.32.1

**Expected behavior**
We expect that no NPE will occur and that an error response will be returned in the “Problem Details” format specified in RFC-7807.

lwitkowski:
Fixes #601

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/httpproblem/security/HttpUnauthorizedUtils.java
+++ b/runtime/src/main/java/io/quarkiverse/httpproblem/security/HttpUnauthorizedUtils.java
@@ -24,10 +24,16 @@ final class HttpUnauthorizedUtils {
                                 .build();
                     }
 
-                    Response.Status status = Response.Status.fromStatusCode(challenge.status);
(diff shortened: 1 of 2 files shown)


pazkooda (approved):
I like the simplicy addressing NPE part of it.

PR #612, opened 2026-05-14
Add per-mapper post-processor registries

modax:
Adds per-mapper `PostProcessorsRegistry` customization by allowing `ExceptionMapperBase` subclasses to supply their own registry. Custom registries can be empty, inherit live from another registry, or use a supplier-backed stream for advanced filtering/composition, while extension defaults remain owned by the built-in global registry.

The idea was briefly explored in https://github.com/quarkiverse/quarkus-http-problem/issues/552#issuecomment-4450781434. This PR does not solve original issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,33 @@ class CustomPostProcessor implements ProblemPostProcessor {
 }
 ```
 
+Custom post-processors registered as CDI beans are applied globally. If a particular mapper needs its own post-processing
+pipeline, pass a custom `PostProcessorsRegistry` to `ExceptionMapperBase`:
+
+```java
+class CustomExceptionMapper extends ExceptionMapperBase<CustomException> {
+
+    private static final PostProcessorsRegistry POST_PROCESSORS = new PostProcessorsRegistry(
+            () -> ExceptionMapperBase.postProcessorsRegistry.stream()
+                    .filter(processor -> !(processor instanceof ProblemLogger)));
+
+    CustomExceptionMapper() {
+        super(POST_PROCESSORS);
+    }
+
+    @Override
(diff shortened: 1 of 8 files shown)


lwitkowski:
@modax thanks for the PR, I really appreciate the effort.

To be frank, I'm not (yet) convinced the use case is wide enough to justify the API surface this introduces.

Two main concerns:
- Scope of the need. The only thing - and please correct me if I'm wrong - this PR seems to enable (beyond what a CDI `ProblemPostProcessor` bean can already do) is excluding or replacing a built-in post-processor for a single mapper. Could you share concrete cases where you need that beyond the `ProblemLogger` example in the README? I'd like to understand whether this is a recurring pattern or a one-off, since it's the first time it's been raised in the history of the extension.

- API shape. This PR makes three previously package-private post-processors public, adds two new `PostProcessorsRegistry` constructors including one exposing `Supplier<Stream<ProblemPostProcessor>>`, and subtly changes `reset()` semantics. That's permanent public surface, and the call-site ergonomics in the README example (`stream().filter(p -> !(p instanceof ProblemLogger))`) suggests the abstraction is leaking implementation details.

Happy to revisit if you can show a concrete use case (or two) that the existing CDI post-processor mechanism can't handle. Otherwise: independent fork may the the most appropriate place for this if you're willing to maintain it.

modax:
Hi, thanks for feedback, very appreciated!

Sorry for staying silent for the last couple of days. I will follow up to this PR eventually, hopefully addressing your concerns.

cescoffier:
This PR would need an "interesting" rebase with the change of architecture.

PR #622, opened 2026-07-02
Make ConstraintViolationExceptionMapper config thread-safe

cescoffier:
Replace two plain static fields with a volatile record reference.

### DIFF
--- a/runtime/src/test/java/io/quarkiverse/httpproblem/validation/ConstraintViolationExceptionMapperTest.java
+++ b/runtime/src/test/java/io/quarkiverse/httpproblem/validation/ConstraintViolationExceptionMapperTest.java
@@ -624,4 +624,70 @@ class ConstraintViolationExceptionMapperTest {
             this.data = data;
         }
     }
+
+    /**
+     * Verifies that configure() and toProblem() are atomic: no request can ever see a status from one
+     * configure() call paired with a title from another configure() call.
+     */
+    @Test
+    void configurationIsAlwaysReadAtomically() throws InterruptedException {
+        final int iterations = 10_000;
+        final ConstraintViolationExceptionMapper testMapper = new ConstraintViolationExceptionMapper();
+        final ConstraintViolationException exception = new ConstraintViolationException("test", null);
+
+        final java.util.concurrent.CountDownLatch startLatch = new java.util.concurrent.CountDownLatch(1);
+        final java.util.concurrent.CountDownLatch doneLatch = new java.util.concurrent.CountDownLatch(2);
+        final java.util.List<HttpValidationProblem> results = new java.util.concurrent.CopyOnWriteArrayList<>();
+
+        // Writer thread: alternates between two consistent (status, title) pairs
+        Thread writer = new Thread(() -> {
+            try {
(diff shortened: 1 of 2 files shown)


lwitkowski:
@cescoffier before I dive into a proper review I would appreciate some context - I'm curious if you encountered this bug yourself or is this simply a good practice for exception handler? Any relation with reactive stack or native mode?

cescoffier:
@lwitkowski It happened to me in a "very fast" dev mode iteration, where it produced a broken outcome. In theory, it could happen in production too, as your recorder may be called after the setup of the HTTP stack during the handling of a request (in Quarkus 4, I've added a build item to prevent this, and also we are working on a better handling of the start-stop sequence - but until that, the race can happen.) 

I'm going to look at this extension a lot more in the next few days/months, so expect more PR from me (working on another issue right now)

lwitkowski:
> @lwitkowski It happened to me in a "very fast" dev mode iteration, where it produced a broken outcome. In theory, it could happen in production too, as your recorder may be called after the setup of the HTTP stack during the handling of a request (in Quarkus 4, I've added a build item to prevent this, and also we are working on a better handling of the start-stop sequence - but until that, the race can happen.)

That makes sense, thanks for explanation, this thread will be useful to our future selves 'why' we did this :)

> I'm going to look at this extension a lot more in the next few days/months, so expect more PR from me (working on another issue right now)

Awesome, looking forward to it.

Once we merge this - do you need a new release or you can iterate on SNAPSHOT built locally?

lwitkowski (approved):

cescoffier:
I don't need a release right now; I need to keep digging in the code and improve a few things (don't worry, nothing big, mostly doc stuff).

runtime/src/test/java/io/quarkiverse/httpproblem/validation/ConstraintViolationExceptionMapperTest.java
```diff
@@ -624,4 +624,70 @@ static class MixedBeanParam {
...
+        // Every result must have a self-consistent (status, title) pair — never a torn read
+        for (HttpValidationProblem problem : results) {
+            int status = problem.getStatusCode();
+            String actualTitle = problem.getTitle();
+            boolean isDefaultPair = status == 400 && "Bad Request".equals(actualTitle);
+            boolean isCustomPair = status == 422 && "Constraint violation".equals(actualTitle);
+            assertThat(isDefaultPair || isCustomPair)
+                    .as("Torn read detected: status=%d title='%s'", status, actualTitle)
+                    .isTrue();
+        }
```
pazkooda, line 691: Maybe it's just me but I find reading this assertion "cumbersome", much more prefer "sentence-like" ones:
```suggestion
        assertThat(results)
                .allMatch(problem -> 
                                (problem.getStatusCode() == 400 && "Bad Request".equals(problem.getTitle())) 
                                        || (problem.getStatusCode() == 422 && "Constraint violation".equals(problem.getTitle())),
                "All results must have a consistent (status, title) pair");
```

With that comment above `for` seems also redundant (aka code should document itself).

PR #623, opened 2026-07-02
Handle malformed and charset-qualified problem+json responses in ThrowingHttpProblemClientExceptionMapper

cescoffier:
Don't expect everyone to return proper JSON.

This PR fixes 2 things:

- The initial code was using MediaType.isCompatible, which is too strict 
- When reading and deserializing the JSON payload, it did not handle failures

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/httpproblem/client/ThrowingHttpProblemClientExceptionMapper.java
+++ b/runtime/src/main/java/io/quarkiverse/httpproblem/client/ThrowingHttpProblemClientExceptionMapper.java
@@ -1,8 +1,10 @@
 package io.quarkiverse.httpproblem.client;
 
+import jakarta.ws.rs.core.MediaType;
 import jakarta.ws.rs.core.Response;
 
 import org.eclipse.microprofile.rest.client.ext.ResponseExceptionMapper;
+import org.jboss.logging.Logger;
 
 import io.quarkiverse.httpproblem.HttpProblem;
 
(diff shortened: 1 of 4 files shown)


cescoffier:
Investigating the CI issue - I'm only using Quarkus REST locally.

runtime/src/main/java/io/quarkiverse/httpproblem/client/ThrowingHttpProblemClientExceptionMapper.java
```diff
@@ -13,18 +15,37 @@
  */
 public class ThrowingHttpProblemClientExceptionMapper implements ResponseExceptionMapper<RuntimeException> {
 
+    private static final Logger log = Logger.getLogger(ThrowingHttpProblemClientExceptionMapper.class);
+
     @Override
     public RuntimeException toThrowable(Response response) {
-        if (!HttpProblem.MEDIA_TYPE.isCompatible(response.getMediaType())) { // TODO add tests here where UTF+8 is appended to respons
```
lwitkowski, line 18: This TODO was here for a reason! :D

lwitkowski: Very nice to see it addressed.

lwitkowski:
Looking very good, yet resteasy tests are failing, they probably throw different exceptions, or at least different messages. Pls have a look.

Apart from this, impl/tests LGTM.

cescoffier:
I've found the issue... Now, I will implement a workaround, but I need to discuss this with @geoand (he is traveling this week) to understand why we have a different default in the Quarkus REST client than in the RESTEasy Classic one (that being said, I believe these Quarkus REST defaults are better).

cescoffier:
Fix pushed (hopefully...)

geoand:
I'll be back tomorrow so I can take a look then

geoand:
@cescoffier what was the difference between Quarkus REST and Quarkus RESTEasy?

lwitkowski:
Thanks a lot @cescoffier !

cescoffier:
@geoand, when the REST client returns a void method, Quarkus REST accepts `*/*`, whereas RESTEasy Classic is much stricter (I couldn't even find what was sent or requested). So the test using RESTEasy Classic failed with a 406 (Unacceptable). Do you know if this difference has been documented somewhere?

geoand:
> Do you know if this difference has been documented somewhere?

I don't think it is, it's probably some edge case that is not covered by the TCK

Issue #586, opened 2026-03-10
Instance field encoding/decoding


Fungrim:
**Describe the bug**
The encoding and decoding of the `instance` field appears to be assymetric and fails when a path contains illegal characters. 

Specifically, the `JacksonProblemSerializer` uses the `InstanceUtils.instanceToPath` method to **decode** the URI before serialization. However, the `JacksonProblemDeserializer` does **not** encode the corresponding string, which results in a `JsonMappingException`. 

Our use case is a REST API (so an exception mapper to serializer) and a REST client with a `ThrowingHttpProblemClientExceptionMapper` (which throws through line 22). 

**To Reproduce**
Here's a simple unit test: 
```
public class UrlEncodingTest {

    @Test
    public void httpProblemInstanceEncoding() throws JsonProcessingException {
        var path = "/api/v1/commons/meeting/666/document/ref/X 1/3";
        var mapper = new ObjectMapper();
        new JacksonProblemModuleRegistrar().customize(mapper);
        var originalProblem = HttpProblem.builder()
            .withTitle("URI Encoding Problem")
            .build();
        var context = ProblemContext.of(null, path);
        var processedProblem = new PostProcessorsRegistry().applyPostProcessing(originalProblem, context);
        var json = mapper.writeValueAsString(processedProblem);
        Assertions.assertDoesNotThrow(() -> {
            mapper.readValue(json, HttpProblem.class);
        });
    }
}
```

**Expected behavior**
I expect a valid HTTP problem which contains a `URI` instance to be serialized and deserialized without exception, the fact that the _unencoded_ path contains illegal characters should not matter. 

I'm happy to provide a PR, but I'm uncertain of the consequences of a straight forward encoding on the client side (or removal of the decoding server side.

Fungrim:
To me it appears that it is the serialization that is wrong. The `type` field is also a URI instance but is encoded when serialized, _and_ I can see existing unit tests checking the de-serialization for illegal URI's.  

EDIT: I can also see the unit test that verifies that the instance _is_ decoded. So... This asymmetry is by design?

pazkooda:
Let me distill the problem. 

You state:
> I expect a valid HTTP problem which contains a **URI** instance to be serialized and **deserialized** without exception, the fact that the *unencoded* path contains **illegal** characters **should not matter**. 

RFC states it should be URI (which by definition needs to conform to given scheme):
[...]

PR #627, opened 2026-07-06
Fix instance field to produce RFC 7807/3986-compliant URI references

cescoffier:
`pathToInstance()` used `URLEncoder.encode()` which encoded the entire path, including `/` separators, producing values like `%2Fapi%2Fusers%2F42`:

- Replaced with `new URI(null, null, path, null)` which encodes only characters invalid in a URI path component (spaces, `|`, `{`, etc.) while preserving `/` as path separator,
- `instanceToPath()` now returns `toASCIIString()` so the serialized `instance` field is a valid [URI reference](https://datatracker.ietf.org/doc/html/rfc7807#section-3.1) per RFC 7807 / RFC 3986

### DIFF
--- a/integration-test/core/src/test/java/io/quarkiverse/httpproblem/GenericMappersIT.java
+++ b/integration-test/core/src/test/java/io/quarkiverse/httpproblem/GenericMappersIT.java
@@ -60,6 +60,6 @@ class GenericMappersIT {
         given()
                 .get("/non|existing path /with unwisecharacters")
                 .then()
-                .body("instance", equalTo("/non|existing path /with unwisecharacters"));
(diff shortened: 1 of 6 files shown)


lwitkowski:
@pazkooda this fix sounds like very related to #586 - maybe you have some input here?

pazkooda:
Will take a look at this tomorrow.

pazkooda:
Looks nearly identical to code I've tried locally when looking on #586 (even tests are very similar).
I'm fine with this as-is ☑️  (just be vary this will "break" backward compatibility of responses - but in a good manner).

lwitkowski:
@pazkooda thanks, I also see it as a bugfix, we can bump minor (not just patch) version for the next release to signal the nature of the change a bit better.

PR #628, opened 2026-07-07
Replace static mutable state with CDI injection

cescoffier:
- Convert PostProcessorsRegistry from a static singleton to a CDI bean.
- Replace CDI.current() lookups and static volatile config in all exception mappers with constructor injection.
- Synthetic beans are now registered during STATIC_INIT so they're available when RESTEasy Classic validates providers.
- No-arg constructors retained for RESTEasy Classic compatibility

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/httpproblem/postprocessing/PostProcessorsRegistry.java
+++ b/runtime/src/main/java/io/quarkiverse/httpproblem/postprocessing/PostProcessorsRegistry.java
@@ -1,45 +1,34 @@
 package io.quarkiverse.httpproblem.postprocessing;
 
-import java.util.List;
-import java.util.concurrent.CopyOnWriteArrayList;
-
-import org.slf4j.LoggerFactory;
-
 import io.quarkiverse.httpproblem.HttpProblem;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Instance;
+import jakarta.inject.Inject;
+
+import java.util.List;
 
 /**
- * Container for prioritised list of Problem post-processors. This class is thread-safe.
+ * Container for prioritised list of Problem post-processors.
+ * Collects all CDI beans implementing ProblemPostProcessor and sorts them by priority at startup.
  */
-public final class PostProcessorsRegistry {
+@ApplicationScoped
+public class PostProcessorsRegistry {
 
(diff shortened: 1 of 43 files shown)


cescoffier:
Opening it as draft, as it's a large change, and I'm not sure we want to go there today (eventually, I think it makes sense).

Basically, the extension was using static mutable instances, which is not really the Quarkus way. I decided (naively, since I thought it would be simple...) to rework the architecture to use CDI and synthesize beans. 

@geoand WDYT?

geoand:
This is the way for sure to make the extension more idiomatic to Quarkus. The principles applied in this PR as the same ones found most extensions in Quarkus core

runtime/src/main/java/io/quarkiverse/httpproblem/postprocessing/PostProcessorsRegistry.java
```diff
@@ -1,45 +1,44 @@
...
     }
 
-    public synchronized void register(ProblemPostProcessor processor) {
-        processors.add(processor);
-        processors.sort(ProblemPostProcessor.DEFAULT_ORDERING);
+    @PostConstruct
+    void init() {
+        processors = processorInstances.stream()
+                .sorted(ProblemPostProcessor.DEFAULT_ORDERING)
+                .collect(Collectors.toUnmodifiableList());
```
geoand: Why is this now needed as the processors are set in the constructor?

cescoffier: You are right it's not needed anymore.

lwitkowski:
I like this direction. 

Btw I think it makes sense to add you guys to the maintainers of this repo. You are obviously much more hands-on with Quarkus nowadays than we (me and @pazkooda) are, and I'm delighted to see your recent contributions. WDYT?

geoand:
Sounds great to me!

cescoffier:
@lwitkowski Sure!

cescoffier:
Making it ready for review.

geoand (approved):

lwitkowski (approved):

lwitkowski:
@cescoffier approved as well, you guys are now maintainers, so feel free to merge anytime (after resolving some conflict, I've merged another PR of yours in the meantime). Great work!

geoand:
Thanks!

@cescoffier this now needs a rebase

PR #629, opened 2026-07-07
Add build-time config to disable specific built-in exception mappers

cescoffier:
Users who want to handle specific exceptions with their own mapper had no way to prevent the built-in one from registering. This adds a per-mapper `enabled` flag keyed by the exception's simple name in kebab-case:

```
quarkus.http-problem.mapper.web-application-exception.enabled=false
quarkus.http-problem.mapper.not-found-exception.enabled=false
```

Disabled mappers are excluded at build time from both RESTEasy classic and reactive registration paths, including the custom reactive mappers for UnauthorizedException and AuthenticationFailedException.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,20 @@ Content-Type: application/problem+json
 }
 ```
 
+- (Build time) Disable specific built-in exception mappers. The key is the exception class simple name in kebab-case.
+
+```
+quarkus.http-problem.mapper.not-found-exception.enabled=false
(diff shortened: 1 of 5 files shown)


lwitkowski:
I'm a bit sceptical with exception name being the key - right now it's unambiguous I believe so it should be fine.
Another option would be to use mapper's name, not the exception it handles, which is fully under control of this extension, but is less user friendly - users think in terms of exceptions I guess.

LGTM

lwitkowski (approved):

PR #634, opened 2026-07-20
Sanitize JSON mapper detail field to prevent leaking internals

### LINKED ISSUE #365: Improve Error Handling with malformed json
Currently Jackson and Jsonb Exception Mappers return responses, which contain detailed information about the malformed request object. From a consumer point of view this makes sense and improves developer expierence.

Example:
`{
  "status": 400,
  "title": "Bad Request",
  "detail": "Internal error: Invalid token=STRING at (line no=3, column no=11, offset=52). Expected tokens are: [COMMA]",
  "instance": "/token"
}`

But the responses also contain implementation details, which from a security point of view is not so good. Especially if your service is a an external public API.
https://owasp.org/www-community/Improper_Error_Handling
Example:
`{
  "status": 400,
  "title": "Bad Request",
  "detail": "Unable to deserialize property 'token' because of: Problem adapting object of type interface org.example.entity.Token to class java.lang.String in class class org.example.enity.TokenTypeAdapter",
  "instance": "/token"
}`

**Solution**

What could help is an option to control the output to return a more generic response and log the detailed error message in the backend.
Example:
`{
  "status": 400,
  "title": "Bad Request",
  "detail": "Syntax Error: malformed json",
  "instance": "/token"
}`

cescoffier:
- Add build-time config property `quarkus.http-problem.include-details` (default false) that controls whether exception mappers expose raw exception messages in the `detail` field
  - When disabled, mappers return "Malformed request body" instead of raw messages that can leak internal class names, adapter names, and implementation structure (Following OWASP Improper Error Handling recommendations)
  - Set quarkus.http-problem.include-details=true to restore the previous behavior for debugging

---

- Fixes: #365

N.B. This is a breaking change in behavior. However, we should not leak implementation details by default.

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/httpproblem/jackson/InvalidFormatExceptionMapper.java
+++ b/runtime/src/main/java/io/quarkiverse/httpproblem/jackson/InvalidFormatExceptionMapper.java
@@ -8,6 +8,8 @@ import jakarta.inject.Inject;
 import jakarta.ws.rs.Priorities;
 import jakarta.ws.rs.core.Response;
 
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.exc.InvalidFormatException;
 
@@ -21,20 +23,30 @@ import io.quarkiverse.httpproblem.postprocessing.PostProcessorsRegistry;
 @Priority(Priorities.USER)
 public final class InvalidFormatExceptionMapper extends ExceptionMapperBase<InvalidFormatException> {
 
+    static final String SANITIZED_DETAIL = "Malformed request body";
+
+    private final boolean includeDetails;
+
     public InvalidFormatExceptionMapper() {
+        this.includeDetails = false;
     }
 
     @Inject
-    public InvalidFormatExceptionMapper(PostProcessorsRegistry postProcessorsRegistry) {
+    public InvalidFormatExceptionMapper(PostProcessorsRegistry postProcessorsRegistry,
+            @ConfigProperty(name = "quarkus.http-problem.include-details", defaultValue = "false") boolean includeDetails) {
         super(postProcessorsRegistry);
+        this.includeDetails = includeDetails;
     }
 
     @Override
     protected HttpProblem toProblem(InvalidFormatException exception) {
+        String detail = includeDetails
+                ? exception.getOriginalMessage()
+                : SANITIZED_DETAIL;
         return HttpProblem.builder()
                 .withStatus(Response.Status.BAD_REQUEST)
                 .withTitle(Response.Status.BAD_REQUEST.getReasonPhrase())
-                .withDetail(exception.getOriginalMessage())
+                .withDetail(detail)
                 .with("field", serializePath(exception.getPath()))
                 .build();
     }
--- a/runtime/src/main/java/io/quarkiverse/httpproblem/jackson/JsonProcessingExceptionMapper.java
+++ b/runtime/src/main/java/io/quarkiverse/httpproblem/jackson/JsonProcessingExceptionMapper.java
@@ -6,6 +6,8 @@ import jakarta.annotation.Priority;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.Priorities;
 
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 
 import io.quarkiverse.httpproblem.ExceptionMapperBase;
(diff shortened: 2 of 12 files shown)


runtime/src/main/java/io/quarkiverse/httpproblem/jackson/InvalidFormatExceptionMapper.java
```diff
@@ -21,20 +22,27 @@
...
 
+    private final boolean includeDetails;
+
     public InvalidFormatExceptionMapper() {
+        this.includeDetails = false;
     }
 
     @Inject
-    public InvalidFormatExceptionMapper(PostProcessorsRegistry postProcessorsRegistry) {
+    public InvalidFormatExceptionMapper(PostProcessorsRegistry postProcessorsRegistry, SanitizeConfig sanitizeConfig) {
```
geoand:
```suggestion
    public InvalidFormatExceptionMapper(PostProcessorsRegistry postProcessorsRegistry, @ConfigProperty("quarkus.http-problem.include-details") includeDetails) {
```

runtime/src/main/java/io/quarkiverse/httpproblem/jackson/JsonProcessingExceptionMapper.java
```diff
@@ -18,16 +19,23 @@
...
 
+    private final boolean includeDetails;
+
     public JsonProcessingExceptionMapper() {
+        this.includeDetails = false;
     }
 
     @Inject
-    public JsonProcessingExceptionMapper(PostProcessorsRegistry postProcessorsRegistry) {
+    public JsonProcessingExceptionMapper(PostProcessorsRegistry postProcessorsRegistry, SanitizeConfig sanitizeConfig) {
```
geoand: Same suggestion as above

runtime/src/main/java/io/quarkiverse/httpproblem/jsonb/JsonbExceptionMapper.java
```diff
@@ -9,21 +9,29 @@
...
 
+    private final boolean includeDetails;
+
     public JsonbExceptionMapper() {
+        this.includeDetails = false;
     }
 
     @Inject
-    public JsonbExceptionMapper(PostProcessorsRegistry postProcessorsRegistry) {
+    public JsonbExceptionMapper(PostProcessorsRegistry postProcessorsRegistry, SanitizeConfig sanitizeConfig) {
```
geoand: Here too

runtime/src/main/java/io/quarkiverse/httpproblem/jsonb/RestEasyClassicJsonbExceptionMapper.java
```diff
@@ -10,17 +10,23 @@
...
+    private final boolean includeDetails;
+
     public RestEasyClassicJsonbExceptionMapper() {
+        this.includeDetails = false;
     }
 
     @Inject
-    public RestEasyClassicJsonbExceptionMapper(PostProcessorsRegistry postProcessorsRegistry) {
+    public RestEasyClassicJsonbExceptionMapper(PostProcessorsRegistry postProcessorsRegistry,
+            SanitizeConfig sanitizeConfig) {
```
geoand: And here

runtime/src/main/java/io/quarkiverse/httpproblem/SanitizeConfig.java
```diff
@@ -0,0 +1,6 @@
+package io.quarkiverse.httpproblem;
+
+public record SanitizeConfig(boolean includeDetails) {
```
geoand: Why have this property when we can just inject config properties where they are needed?

runtime/src/main/java/io/quarkiverse/httpproblem/jackson/InvalidFormatExceptionMapper.java
cescoffier: I agree, I was wondering. In my first attempt I had multiple fields. But now, with a single one, we can drop the record.

runtime/src/main/java/io/quarkiverse/httpproblem/SanitizeConfig.java
cescoffier: Initially I had a more fine grain support (per exception type). I the. I removed it, I found it too complex. That's why I added this record.

geoand: FWIW, Quarkus config interfaces can also be injected

cescoffier: I should use this for the other config objects.

geoand (approved):
Great!

lwitkowski (approved):

PR #638, opened 2026-07-26
Publish core integration tests for Quarkus Platform

lwitkowski:
Always include the core IT module in the Maven reactor (not only under -Pit) and attach a test-jar so releases publish both the main artifact and *-tests.jar. Optional IT modules (openapi, zalando, rest-client) stay behind the it profile for this repo's CI matrix.

Required so Quarkus Platform can re-run the suite for BOM compatibility: https://github.com/quarkusio/quarkus-platform/pull/1982 https://github.com/quarkusio/quarkus-platform/pull/1982#issuecomment-4874584007. Platform tests will exercise the reactive REST + Jackson path only — not the full CI matrix (jsonb, classic rest providers). That is usually fine as the representative suite.

The main reason it was **not** published before is [official quarkiverse faq](https://github.com/quarkiverse/quarkiverse/wiki/Release#release-fails-while-deploying-the-integration-tests-and-docs) entry, which is not really applicable to Platform extensions.

### DIFF
--- a/integration-test/core/pom.xml
+++ b/integration-test/core/pom.xml
@@ -22,4 +22,20 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <!-- Required so Quarkus Platform can re-run these ITs from the published *-tests.jar -->
+            <plugin>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>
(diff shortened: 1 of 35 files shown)


gsmet (commented):
@lwitkowski :wave: thanks for the PR! I have been tasked to make sure we include quarkus-http-problem in the Platform soon so glad you took this one.

Is there a reason why we don't also publish the `openapi` and `rest-client` modules? They look like something we want to make sure it works?

gsmet:
And maybe we should also publish the module added by @phillip-kruger with the TCK tests?

gsmet:
FWIW, I experimented a bit with this and the fact that the test dependencies are in the parent POM of the ITs is problematic for the Platform generator.

Same for the profiles. I'll dig more and get back to you.

aloubyansky:
I'm fixing the dependencies in the parent issue. It will make sense to fix whether you change it in this project or not.

gsmet:
@lwitkowski hey. So I ended up reworking the test infra a bit to use Surefire for the `@QuarkusTest` and only use Failsafe for the native ITs.

I ran quite a lot of tests so I hope CI will be happy.

I also publish all the ITs, not just Core and then we will pick and choose in the Platform.

Let me know if you are happy with these changes as I would like to get them merged and released before tomorrow afternoon so that I can include Quarkus HTTP Problem in the 3.38 Platform.

lwitkowski:
All looks good @gsmet

gsmet:
Thanks, let's merge then :).

PR #640, opened 2026-07-27
Add Jackson exception mappers for MismatchedInputException and InvalidDefinitionException

### LINKED ISSUE #317: Not all Jackson exceptions mapper have been overruled
Quarkus: 3.4.3
Resteasy-Problem: 3.0.0

It seems like that not all ``Jackson`` exceptions are handled by `resteasy-problem`.

image

Although there is a mapper for ``JsonProcessingException``, both child exceptions ``MismatchedInputException`` and ``InvalidDefinitionException`` are handled by Jackson itself. 

According to the spec, this seems totally fine:

```
When choosing an exception mapping provider to map an exception, an implementation MUST use the
provider whose generic type is the nearest superclass of the exception. If two or more exception providers
are applicable, the one with the highest priority MUST be chosen as described in Section 4.1.3.
```

That is a little annoying cause it looks like you need to overrule each exception mapper individually. 

Do you have any strategy in place to detected new mappers automatically?

phillip-kruger:
Two `JsonMappingException` subclasses - `MismatchedInputException` and `InvalidDefinitionException` - were not covered by the extension, so they fell through to Jackson's own mappers and bypassed the http-problem pipeline (logging, MDC injection, detail sanitization, RFC 9457 response format).

This adds dedicated mappers for both exception types following the same pattern as `InvalidFormatExceptionMapper`, including field path extraction and the `include-details` sanitization config. It also extracts the shared field path serialization logic into a `JacksonFieldPathSerializer` utility to avoid duplication across the three mappers that need it.

Both mappers can be individually disabled via `quarkus.http-problem.mapper.mismatched-input-exception.enabled=false` and `quarkus.http-problem.mapper.invalid-definition-exception.enabled=false`.

Fixes #317

### DIFF
--- a/runtime/src/main/java/io/quarkiverse/httpproblem/jackson/InvalidDefinitionExceptionMapper.java
+++ b/runtime/src/main/java/io/quarkiverse/httpproblem/jackson/InvalidDefinitionExceptionMapper.java
@@ -5,7 +5,7 @@ import jakarta.inject.Inject;
 import jakarta.ws.rs.Priorities;
 import jakarta.ws.rs.core.Response;
 
-import com.fasterxml.jackson.databind.exc.InvalidFormatException;
+import com.fasterxml.jackson.databind.exc.InvalidDefinitionException;
 
 import io.quarkiverse.httpproblem.DetailSanitizer;
 import io.quarkiverse.httpproblem.ExceptionMapperBase;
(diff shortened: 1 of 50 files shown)


gsmet:
@phillip-kruger I rebased this one on top of @cescoffier 's work and added a small commit to add a comment and use a consistent priority.

Let me know what you think.

@lwitkowski I think we will also want this one in the first release we include in the Platform.

runtime/src/main/java/io/quarkiverse/httpproblem/jackson/JacksonFieldPathSerializer.java
```diff
@@ -0,0 +1,37 @@
+package io.quarkiverse.httpproblem.jackson;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.databind.JsonMappingException;
+
+final class JacksonFieldPathSerializer {
```
lwitkowski, line 8: nit: this could use targeted unit test

runtime/src/main/java/io/quarkiverse/httpproblem/jackson/InvalidDefinitionExceptionMapper.java
```diff
@@ -0,0 +1,49 @@
...
+            ProblemRuntimeFixedConfig config) {
+        super(postProcessorsRegistry);
+        this.includeDetails = config.includeDetails();
+    }
+
+    @Override
+    protected HttpProblem toProblem(InvalidDefinitionException exception) {
+        String detail = includeDetails
+                ? exception.getOriginalMessage()
+                : SANITIZED_DETAIL;
```
lwitkowski: nit: this looks innocent, but we have the same `logic` (trivial but still logic) in few other places now, I'd probably wrap this into simple bean that would take care of this sanitisation based on configuration.

lwitkowski (approved):

lwitkowski:
@phillip-kruger added 2 small optional nits, feel free to merge as is, up to you.

@gsmet I feel like I'm the bottleneck in some PRs, I appreciate you try to keep me in the loop, but I fully trust your judgement, so feel free to move things forward if I'm not answering in timely manner :)

gsmet:
@lwitkowski of course, I keep you in the loop :).

From what I can see, @phillip-kruger addressed your comments so we could merge!

gsmet:
@phillip-kruger I don't know what you did there but you reverted a ton of changes including changes from other pull requests that were merged before this one...

I'm going to fix this mess but please be more careful...

phillip-kruger:
@gsmet - sorry about that, I must have rebased on an old branch .

PR #641, opened 2026-07-27
Add configurable log levels and stack trace thresholds

### LINKED ISSUE #512: Configuration option for choosing at for which status codes to log stacktrace
This is not related to RFC7807

**Description of problem**
In some of our services we would very much like to always log the stack trace whenever something goes wrong such as a 403-Forbidden. 

However the current setup for the ProblemLogger only logs stack traces if status code is 500 or above. [Problem Logger](https://github.com/quarkiverse/quarkus-resteasy-problem/blob/main/runtime/src/main/java/io/quarkiverse/resteasy/problem/postprocessing/ProblemLogger.java#L31)

Something simlar to this would be nice to be able to control at which cutoff point you start logging info instead of error:
```
/**
 * Logs problems with ERROR (for HTTP with status above threshold default = 500) or INFO (other exceptions) log level.
 * In case of ERROR stack trace is printed as well.
 */
final class ProblemLogger implements ProblemPostProcessor {

    private final Logger logger;

    ProblemLogger(Logger logger) {
        this.logger = logger;
    }

    @ConfigProperty(name = "quarkus.resteasy.problem.status-code-threshold", defaultValue = "500")
    int statusCodeThreshold;

    @Override
    public HttpProblem apply(HttpProblem problem, ProblemContext context) {
        if (!logger.isErrorEnabled()) {
            return problem;
        }

        if (problem.getStatusCode() >= statusCodeThreshold) {
            if (logger.isErrorEnabled()) {
                logger.error(serialize(problem), context.cause);
            }
        } else {
            if (logger.isInfoEnabled()) {
                logger.info(serialize(problem));
            }
        }
        return problem;
    }

   // ... more code - that was cut out for simplicity
}
```

### LINKED ISSUE #552: Allow customization of log levels
No.

Currently, the `ProblemLogger` logs everything to log level `INFO` when the response is `< 500` and everything `>=500` on level `ERROR`.

It would be nice to:
- have a way to turn off logging completely (without setting the `min-level` for `http-problem` to `FATAL`),
- be able to define a default log level, and
- to be able to define a log level per response code and/or response code range (e.g. `401` and `403` on level `WARN`, everything else `< 500` on level `INFO`, everything `>= 500` on level `WARN`)

**Additional information**
This is slightly related to https://github.com/quarkiverse/quarkus-resteasy-problem/issues/512, but https://github.com/quarkiverse/quarkus-resteasy-problem/issues/512 focuses on stack traces, while this issue focuses on the log level.

phillip-kruger:
The ProblemLogger previously hardcoded its behavior - 4xx at INFO without stack traces, 5xx at ERROR with stack traces. This makes both the log level and stack trace inclusion configurable per status code class or exact code, via build-time configuration:

```properties
quarkus.http-problem.logging.enabled=true
quarkus.http-problem.logging.level.4xx=INFO
quarkus.http-problem.logging.level.5xx=ERROR
quarkus.http-problem.logging.level.401=WARN
quarkus.http-problem.logging.include-stack-trace=5xx,403
```

Exact codes take precedence over status classes. Logging can be disabled entirely with `logging.enabled=false`. Defaults preserve the existing behavior (4xx=INFO, 5xx=ERROR, stack traces for 5xx only), so this is fully backwards compatible.

ProblemLogger is now created as a synthetic bean via the recorder (following the MdcPropertiesInjector pattern) so it can receive the resolved configuration at build time.

Fixes #552
Fixes #512

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/httpproblem/deployment/ProblemBuildConfig.java
+++ b/deployment/src/main/java/io/quarkiverse/httpproblem/deployment/ProblemBuildConfig.java
@@ -1,7 +1,9 @@
 package io.quarkiverse.httpproblem.deployment;
 
+import java.util.List;
 import java.util.Map;
 
+import io.quarkiverse.httpproblem.postprocessing.ProblemLogLevel;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
 import io.smallrye.config.ConfigMapping;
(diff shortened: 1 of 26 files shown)


gsmet:
This will need a rebase because we merged @cescoffier 's work.

gsmet (approved):

PR #652, opened 2026-07-29
Replace ClasspathDetector subclasses with Quarkus Capabilities API

### LINKED ISSUE #651: Replace ClasspathDetector Class.forName() probes with Capabilities API
The `ClasspathDetector` hierarchy (`RestEasyClassicDetector`, `RestEasyReactiveDetector`, `JacksonDetector`, `JsonBDetector`, `OpenApiDetector`) uses `Class.forName()` to probe for implementation-internal Quarkus classes to decide which build steps and exception mappers to register.

This is fragile - the probed classes (e.g. `ResteasyContextProvider`, `QuarkusContextProducers`, `ObjectMapperCustomizer`) are internal and could be renamed or moved between Quarkus versions.

Quarkus provides the `Capabilities` API with stable constants (`Capability.RESTEASY`, `Capability.JACKSON`, `Capability.JSONB`, etc.) for exactly this purpose. The extension already uses `Capabilities` in `ProblemProcessor.createFeature()` for JSON provider detection, so the dependency is in place.

The `@BuildStep(onlyIf = ...)` methods would need to be refactored to receive `Capabilities` as a parameter and do the check in the method body, rather than relying on `BooleanSupplier` conditions. The `ExceptionMapperDefinition.onlyIf()` filtering inside `neededExceptionMappers()` would similarly need to take `Capabilities` as context.

Note: the base `ClasspathDetector` usage in `ExceptionMapperDefinition.thatHandles()` (which checks whether the *exception class itself* exists) is still valid and should be kept.

phillip-kruger:
The five `ClasspathDetector` subclasses (`RestEasyClassicDetector`, `RestEasyReactiveDetector`, `JacksonDetector`, `JsonBDetector`, `OpenApiDetector`) use `Class.forName()` to probe for internal Quarkus implementation classes. These internal classes can be renamed or moved between Quarkus versions, making the detection fragile.

This replaces them with the stable `Capabilities` API that Quarkus provides for exactly this purpose, using constants like `Capability.RESTEASY`, `Capability.RESTEASY_REACTIVE`, `Capability.JACKSON`, `Capability.JSONB`, and `Capability.SMALLRYE_OPENAPI`.

The `@BuildStep(onlyIf = ...)` annotations are replaced with `Capabilities` parameter injection and early-return checks in the method body. The `ExceptionMapperDefinition.onlyIf()` method now takes a capability string instead of a `BooleanSupplier`. The base `ClasspathDetector` is kept for checking whether exception classes themselves exist on the classpath, which is still a valid use case.

- Closes: #651

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/httpproblem/deployment/ExceptionMapperDefinition.java
+++ b/deployment/src/main/java/io/quarkiverse/httpproblem/deployment/ExceptionMapperDefinition.java
@@ -1,7 +1,8 @@
 package io.quarkiverse.httpproblem.deployment;
 
 import java.util.Objects;
-import java.util.function.BooleanSupplier;
+
+import io.quarkus.deployment.Capabilities;
 
 final class ExceptionMapperDefinition {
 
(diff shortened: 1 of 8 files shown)


geoand (approved):
Makes sense!

